PR #3, opened 2018-10-16
Fix Django version to avoid security vulnerability.

wangxf123456:


wangxf123456:
Fix Django version to avoid security vulnerability.

PR #10, opened 2018-10-17
Refactor project creation workflow.

wangxf123456:


django_cloud_deploy/integration_tests/workflow_test.py
```diff
@@ -246,3 +248,49 @@ def test_deploy_new_app_sync(self):
...
+        super().setUp()
+        self._project_workflow = _project.ProjectWorkflow(self.credentials)
+
+    def test_create_new_project_no_permission(self):
+        project_name = 'New Project'
+        project_id = utils.get_resource_name(resource_type='project')
+
+        exception_regex = r'.*HttpError 403.*'
+
+        # The provided credentials object does not have permission to create
```
enriquejosepadilla: Why does this credential not work but in the other tests it does work?

django_cloud_deploy/integration_tests/workflow_test.py
```diff
@@ -246,3 +248,49 @@ def test_deploy_new_app_sync(self):
                     secrets=secrets)
                 response = requests.get(admin_url)
                 self.assertIn('Django administration', response.text)
+
+
+class ProjectWorkflowIntegrationTest(test_base.BaseTest):
```
enriquejosepadilla: Theres no test for positive creation?

django_cloud_deploy/workflow/_project.py
```diff
@@ -0,0 +1,71 @@
...
+"""Manages Google Cloud Platform projects creation."""
+
+import enum
+
+from django_cloud_deploy.cloudlib import project
+
+from google.auth import credentials
+
+
+class CreationMode(enum.Enum):
```
enriquejosepadilla: Why do we need CreationMode? Don't we always want to create it ? Is this for tests to not create an already exisiting project? Not sure why MUST_EXIST is needed.

django_cloud_deploy/integration_tests/workflow_test.py
wangxf123456: This credentials object only allow users to manage resources under a pre-created GCP project. Other tests can do whatever they need to do under that project, but they do not have access to other projects, including creation of other projects.

django_cloud_deploy/integration_tests/workflow_test.py
wangxf123456: Yeah. The way our integration test works does not allow creation of new GCP projects,

django_cloud_deploy/workflow/_project.py
wangxf123456: Because we cannot create new projects, this is useful for our integration test.

django_cloud_deploy/workflow/_project.py
```diff
@@ -0,0 +1,71 @@
...
+            project_name: Name of the GCP project the caller wants to create.
+            project_id: Id of the GCP project to create.
+            project_creation: Whether we want to create the GCP project or use
+                an existing project.
+
+        Returns:
+            The project id of the created project.
+
+        Raises:
+            ValueError: If we want to create the project and the project
```
brianquinlan: I think that you are using custom exceptions in other modules e.g. _deploygke.py. Maybe ProjectionCreationError?

django_cloud_deploy/workflow/_project.py
brianquinlan: It also might be useful if we let the user add a Django cluster to an existing project.

enriquejosepadilla: So we would want to ask the user if they would like to create a new project or use an existing one?

If they use an exisiting one they have to choose from a list or pass in the ID?

And for a new one could follow the exisiting flow.

PR #11, opened 2018-10-17
Modify setup.py to to install the right requirements.

enriquejosepadilla:
Fixed a typo and also pinned urllib3.


setup.py
```diff
@@ -18,6 +18,7 @@
   long_description = fh.read()
 
 install_requires=[
+  'urllib3==1.23', #https://github.com/requests/requests/issues/4830
```
brianquinlan: Do you want to bump the version number in the same change?

enriquejosepadilla: for some reason or another it didn't pick up the version change -_-, It should of.

brianquinlan: I mean the version of django-cloud-deploy

PR #15, opened 2018-10-19
Static content serve refactor

wangxf123456:
Add static content serve workflow and fix cloudlib/static_content_serve.StaticContentServeClient.make_bucket_public.

Previously that method does not really the bucket publicly available.


django_cloud_deploy/nox.py
```diff
@@ -61,4 +62,4 @@ def integration_test(session, python_version):
 
     session.interpreter = 'python{}'.format(python_version)
     session.install(*PACKAGES)
-    session.run('pytest', 'integration_tests')
+    session.run('py.test', 'integration_tests', '--forked')
```
wangxf123456: This will make each integration test case run in a separate process. Our integration tests generate several Django projects. Because Django settings can only be configured once, all test cases will target at the Django source files created the first time. This causes some tests to fail.

Issue #20, opened 2018-10-23
Use tool for existing GCP project.


enriquejosepadilla:
Will write comments with questions regarding possible bugs.

enriquejosepadilla:
How many of our steps could be affected by this? 

Off the top of my head I'm thinking enabling API's **could** throw an error if the API's are already enabled?

enriquejosepadilla:
It was discussed that this is an edge case due to our testing and will not be a supported feature.

PR #22, opened 2018-10-25
Allow tool to take allow users to use an existing GCP project.

enriquejosepadilla:
https://github.com/GoogleCloudPlatform/django-cloud-deploy/issues/20


django_cloud_deploy/cli/new.py
```diff
@@ -118,13 +124,17 @@ def main(args: argparse.Namespace):
     if remaining_parameters_to_prompt:
```
enriquejosepadilla: This has the an issue if they want to use an exisiting project but they pass in all the arguments. The check to confirm that the name is correct wouldn't happen.

enriquejosepadilla:
#20

PR #24, opened 2018-10-25
Add workflow manager.

wangxf123456:


django_cloud_deploy/cli/new.py
```diff
@@ -87,6 +94,7 @@ def main(args: argparse.Namespace):
         'django_superuser_login',
         'django_superuser_password',
         'django_superuser_email',
+        'billing_account_name'
```
enriquejosepadilla: Do we want this at the end?

django_cloud_deploy/workflow/__init__.py
```diff
@@ -11,3 +11,160 @@
...
+
+        print(self._generate_section_header(8, 'Deployment'))
+        admin_url = self._deploygke_workflow.deploy_new_app_sync(
+            project_id, cluster_name, django_directory_path,
+            django_project_name, image_name, secrets)
+        if open_browser:
+            webbrowser.open(admin_url)
+
+    def _generate_section_header(self, step: str, section_name: str):
+        return '\n\n**Step {} of {}: {}**\n\n'.format(
```
enriquejosepadilla: Just use one \n. It looks way too large of a difference.

PR #41, opened 2018-11-01
Make workflow manager able to accept customized service accounts and services.

wangxf123456:
This is to make it ready for e2e test. All e2e tests are running under the same GCP project. So we need to create service accounts with random ids.

I put the logic of creating Kubernetes secret from service_accout_workflow to workflow manager. 
After this change, we can still create several service accounts by modifying workflow/data/service_accounts.json.


django_cloud_deploy/workflow/__init__.py
```diff
@@ -82,6 +87,18 @@ def create_and_deploy_new_project(self,
             django_directory_path: The location where the generated Django
                 project code should be stored.
             database_password: The password for the default database user.
+            required_services: The services need to enable for deployment.
```
enriquejosepadilla: need to enable, need to be enabled*

django_cloud_deploy/workflow/__init__.py
```diff
@@ -82,6 +87,18 @@ def create_and_deploy_new_project(self,
...
+            required_service_accounts: Service accounts need to create for
```
enriquejosepadilla: need to be created*

django_cloud_deploy/workflow/__init__.py
```diff
@@ -151,9 +169,24 @@ def create_and_deploy_new_project(self,
             self._generate_section_header(
                 7, 'Create Service Account Necessary For Deployment',
                 self._TOTAL_NEW_STEPS))
-        secrets = self._service_account_workflow.handle_service_accounts(
-            project_id, database_username, database_password
-        )
+
+        # Generate required Kubernetes secrets with service account keys.
+        secrets = {
```
enriquejosepadilla: can we put all of this logic into one function? generate_secrets?

django_cloud_deploy/workflow/_service_account.py
```diff
@@ -72,105 +30,32 @@ def __init__(self, credentials: credentials.Credentials):
...
-
-        return service_account_keys
-
-    def create_key(self, project_id: str, service_account_obj: ServiceAccount):
-        """Creates the service accounts and retrieves the key for the project.
-
-        Args:
-            project_id: The GCP project id to create service accounts.
-            service_account_obj: The service account to create and retrieve key.
+            The service account key content.
```
brianquinlan: Could you show an example return value?

django_cloud_deploy/workflow/__init__.py
```diff
@@ -226,3 +247,53 @@ def _generate_section_header(self, step: str, section_name: str,
...
+        """Generate Kubernetes secrets required for deployment.
+
+        Args:
+            project_id: The unique id for your Google Cloud Platform project.
+            database_username: Name of the default database user.
+            database_password: The password for the default database user.
+            required_service_accounts: Service accounts needed by deployment.
+
+        Returns:
+            All secrets necessary for deployment.
```
brianquinlan: Could you provide an example?

wangxf123456: Example added.

django_cloud_deploy/workflow/__init__.py
wangxf123456: Done.

django_cloud_deploy/workflow/__init__.py
wangxf123456: Done.

django_cloud_deploy/workflow/__init__.py
wangxf123456: Done.

PR #46, opened 2018-11-03
Enforce google style in presubmit. Fixes #2.

### LINKED ISSUE #2: Suggest "yapf" in contributing documentation and create a reasonable default config
See: https://github.com/google/yapf#formatting-style

brianquinlan:


django_cloud_deploy/nox.py
```diff
@@ -51,8 +51,8 @@ def unit_test(session, python_version):
 def lint(session):
     """Run linters."""
     session.interpreter = 'python3.5'
-    session.install('flake8')
-    session.run('flake8')
+    session.install('yapf')
+    session.run('yapf', '--diff', '-r', 'cli', 'cloudlib', 'workflow')
```
wangxf123456: Should we also check skeleton/ and tests/?

brianquinlan: I think that we should do tests (added) but skeleton should probably follow whatever style Django uses.

wangxf123456: I see.

PR #50, opened 2018-11-06
Modify the return type of ServiceAccountClient.create_key

wangxf123456:


wangxf123456:
This is an existing bug. In workflow/_service_account.py, it directly returns the return value of ServiceAccountClient.create_key, which is expected to be a string.

PR #51, opened 2018-11-06
Add type annotations for skeleton/source_generator.py.

wangxf123456:


django_cloud_deploy/skeleton/source_generator.py
```diff
@@ -41,7 +41,10 @@ class _Jinja2FileGenerator(_FileGenerator):
     def __init__(self):
         self._template_env = jinja2.Environment()
 
-    def _render_file(self, template_path, output_path, **options):
+    def _render_file(self,
+                     template_path: str,
+                     output_path: str,
+                     **options: Any):
```
enriquejosepadilla: isn't options a Dict[str, Any] ?

wangxf123456: I changed the type of it.

PR #54, opened 2018-11-06
Refactor of skeleton/source_generator.py

wangxf123456:
Separate settings file generation from Django file generation. 

To deploy an existing app, we may not need to generate Django project files, but we will always need to generate new settings file based on existing settings files. So we should move those that out.


django_cloud_deploy/skeleton/source_generator.py
```diff
@@ -110,35 +110,18 @@ class _DjangoFileGenerator(_Jinja2FileGenerator):
     ADMIN_TEMPLATE_FOLDER = 'admin_template'
     PROJECT_TEMPLATE_FOLDER = 'project_template'
 
-    def generate_project_files(self,
-                               project_id: str,
-                               site: str,
-                               destination: str,
-                               database_name: Optional[str] = None,
-                               cloud_storage_bucket_name: Optional[str] = None):
+    def generate_project_files(self, site: str, destination: str,):
```
brianquinlan: trailing comma

PR #61, opened 2018-11-07
Run pytype for type checking.

brianquinlan:
I mean this to be run manually with nox -s type_check


wangxf123456:
Should we add another label showing the status of pytype checks?

brianquinlan:
Maybe, what does adding a new label do? Keep in mind that the status will always be green because I'm ignoring the return code.

PR #62, opened 2018-11-08
Add another command for users to setup docker.

wangxf123456:
Some users may not have the "docker" group in their system. We need to create it first.


django_cloud_deploy/tool_requirements.py
```diff
@@ -161,8 +161,9 @@ def check(cls):
             # TODO: Check output for error message given when user has
             # not ran the command
             msg = ('Docker is installed but not correctly set up.'
-                   'Use the following command to fix it: '
-                   'sudo usermod -a -G docker $USER\n'
+                   'Use the following command to fix it: \n'
+                   '$sudo groupadd docker\n'
```
brianquinlan: Add a space after the "$"s.

wangxf123456: Done.

PR #65, opened 2018-11-09
 Make source_generator able to generate settings file based on existing settings files.

wangxf123456:
Right now this can only handle the simplest cases. There is expected to have only one settings file with name settings.py.


django_cloud_deploy/skeleton/source_generator.py
```diff
@@ -110,6 +116,19 @@ class _DjangoFileGenerator(_Jinja2FileGenerator):
     ADMIN_TEMPLATE_FOLDER = 'admin_template'
     PROJECT_TEMPLATE_FOLDER = 'project_template'
 
+    @staticmethod
+    def exist(project_dir: str, project_name: str) -> bool:
+        """Returns whether the directory contains Django fils."""
```
enriquejosepadilla: s/fils/files.

django_cloud_deploy/skeleton/source_generator.py
```diff
@@ -412,3 +496,58 @@ def generate_all_source_files(self,
...
+            image_tag: A customized docker image tag used in integration tests.
+        """
+        destination = os.path.abspath(os.path.expanduser(destination))
+        self.settings_file_generator.generate_from_existing(
+            project_id, project_name, destination, database_name,
+            cloud_storage_bucket_name)
+
+        self.docker_file_generator.generate(project_name, destination)
+
+        # TODO: Generate requirements.txt based on existing file.
```
enriquejosepadilla: please give priority to this one, as it can definitely break things for user.

enriquejosepadilla:
Looks good, will approve when tests are passing.

wangxf123456:
The tests were just flaky. I rerun the tests and then they pass.

PR #75, opened 2018-11-15
Add a better description to DjangoFilesystemPath.

brianquinlan:


django_cloud_deploy/cli/prompt.py
```diff
@@ -525,8 +525,8 @@ def prompt(cls,
 
         while True:
             console.tell(
-                ('{} Enter the directory of the Django project you want to '
-                 'update:'.format(step_prompt)))
+                ('{} Enter a new directory path to store project source, '
```
wangxf123456: I think this change should be made on line 481. This class is only used for update command.

brianquinlan: Oops. Fixed.

PR #82, opened 2018-11-21
Refactor source_generator.py.

wangxf123456:
Two functions are added to source generators.

The function "generated" will return whether the source files are already generated in the given directory.

The function "exist" will return whether the directory already contains the source file we need to generate. Note that the source file might not be generated by us. For example, a directory may already contain a "settings.py", but we still need to generate "local_settings.py", "base_settings.py"and "remote_settings.py" based on that.

The new workflow for each generator is like the follows:
```python
if generated():
    return

if exist():
  generate_from_existing()
else:
  generate_new()
```

Also some lint errors are fixed.


django_cloud_deploy/skeleton/source_generator.py
```diff
@@ -337,95 +442,150 @@ def generate_from_existing(self,
...
+        for file_name in _AppEngineFileGenerator._FILES:
+            if file_name not in files_list:
+                return False
+        return True
+
     # TODO: remove database_password
-    def generate(self, project_name: str, destination: str,
+    def generate(self, project_name: str, project_dir: str,
                  database_password: str):
         """Generate Dockerfile and .dockerignore.
```
enriquejosepadilla: Meh copy paste error from my PR. If you can fix it ill appreciate it, else I'll do it in my next PR. :)

enriquejosepadilla: Should be App.yaml + .dockerignore.

PR #86, opened 2018-11-26
Make update command ask for less information from users.

wangxf123456:


django_cloud_deploy/skeleton/source_generator.py
```diff
@@ -747,6 +747,8 @@ def generate_all_source_files(self,
                                            database_name,
                                            cloud_storage_bucket_name,
                                            cloud_sql_connection_string)
+        self.app_engine_file_generator.generate(project_name, project_dir,
```
enriquejosepadilla: this was added in the other PR, when was this removed?

wangxf123456: It was deleted by accident in https://github.com/GoogleCloudPlatform/django-cloud-deploy/pull/82/files

PR #89, opened 2018-11-27
Make it so the default app shows a hello world at the root url.

brianquinlan:


wangxf123456:
We should also modify django_cloud_deploy/tests/lib/test_base.py, which calls source_generator.generate_all_source_files().

PR #91, opened 2018-11-27
App engine clean up

enriquejosepadilla:


django_cloud_deploy/skeleton/templates/settings_template/remote_settings.py-tpl
```diff
@@ -11,9 +11,18 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import json
+from google.cloud import storage
 
 from .base_settings import *
 
+def get_database_password():
```
wangxf123456: I am not feeling good about putting this function in a settings file. But I do not know where to put this logic. Maybe we should ask @brianquinlan

enriquejosepadilla: I think there is no doubt that this is the wrong place to put it. All of us have thought the same thing. I asked @brianquinlan, and we agreed to put it there until somewhere else would make more sense.

brianquinlan: I can't think of a better place for this. But could we use https://cloud.google.com/kms/docs/secret-management ?

I would definitely make the function private.

Issue #78, opened 2018-11-16
Make the tool need only one authentication step


enriquejosepadilla:
Some participants were frustrated/confused about logging in during back to back steps as they perceived it to be redundant

We believe creating two steps which clearly define what auth is doing will be a better experience. This is a smaller commitment to figuring out how to only do one auth screen (if its even possible)

@wangxf123456 talked to offline, he agrees.
@brianquinlan What do you think?

PR #95, opened 2018-11-27
Make the tool need only one authentication step

wangxf123456:
Previously pushing to container registry requires "gcloud login". So we use another way to authenticate docker.


wangxf123456:
Fixes #78

PR #97, opened 2018-11-28
Create a new getting started guide

brianquinlan:
The intention is that this should provide complete getting started documentation. Please let me know if I should add something.


GETTING-STARTED.md
```diff
@@ -0,0 +1,110 @@
...
+## Setup and Installation
+
+### Prerequisites
+
+In order to use **Django Deploy**, you must first install the following dependencies:
+- [Python](https://www.python.org/downloads/) 3.5 or higher
+- [virtualenv](https://virtualenv.pypa.io/en/stable/installation/)
+- [Docker](https://docs.docker.com/install/overview/) (any edition)
+- [Google Cloud SDK](https://cloud.google.com/sdk/docs/quickstarts)
+- [kubectl](https://cloud.google.com/kubernetes-engine/docs/quickstart) (can be
```
wangxf123456: kubectl is not needed anymore.

GETTING-STARTED.md
```diff
@@ -0,0 +1,110 @@
+# Getting Started
+
+## Setup and Installation
+
+### Prerequisites
+
+In order to use **Django Deploy**, you must first install the following dependencies:
+- [Python](https://www.python.org/downloads/) 3.5 or higher
```
enriquejosepadilla: Should we tell users to check with python --version ?

GETTING-STARTED.md
```diff
@@ -0,0 +1,110 @@
...
+[Setup and Installation steps](#Setup-and-Installation) and running this
+command:
+```bash
+$ django-cloud-deploy new
+```
+
+Follow the prompts displayed in the terminal. Make sure that you remember the
+following information:
+
+ - the default database password
```
enriquejosepadilla: Is the password a default?

GETTING-STARTED.md
brianquinlan: Removed.

GETTING-STARTED.md
brianquinlan: I added a note in the installation section but I'm not super happy with how I did it...

GETTING-STARTED.md
brianquinlan: Update the phrasing.

GETTING-STARTED.md
enriquejosepadilla: i think its clear on our github about 3.5, at this point if that fails for them then GG, else we can add another check to make sure their python and fail fasT?

PR #111, opened 2018-11-29
Add a function to get project name given a django directory.

wangxf123456:
This is helpful when deploying an existing app. We do not need to ask users their project name anymore.


django_cloud_deploy/skeleton/utils.py
```diff
@@ -0,0 +1,54 @@
...
+    "os.environ.setdefault('DJANGO_SETTINGS_MODULE',
+                           '{{ project_name }}.settings')"
+
+    We can determine the project name from this line.
+
+    Args:
+        django_directory_path: Absolute path of django project directory.
+
+    Raises:
+        ProjectContentError: When the function cannot find Django project name.
```
enriquejosepadilla: This can raise another error, if on line 39 doesn't have manage.py in the path.

django_cloud_deploy/tests/unit/skeleton/utils_test.py
```diff
@@ -0,0 +1,33 @@
...
+import shutil
+import tempfile
+import unittest
+
+from django.core import management
+
+from django_cloud_deploy.skeleton import utils
+
+
+class UtilTest(unittest.TestCase):
```
enriquejosepadilla: Can you add a test where the file doesnt exist?

django_cloud_deploy/skeleton/utils.py
wangxf123456: Done.

django_cloud_deploy/tests/unit/skeleton/utils_test.py
wangxf123456: Done.

PR #115, opened 2018-12-01
Add extra_settings.py to save our customized settings attributes.

wangxf123456:
This is for deploying an existing app. To deploy an existing app, we need to find some ways to merge our settings file and the existing settings file. When deploying an existing app, what we do is as the follows:

1. Move existing settings file to 'base_settings.py'
2. Create 'extra_settings.py' which inherits 'base_settings.py'. 'extra_settings.py' only includes our customization of Django apps. Now our settings and existing settings are merged.
3. Create 'local_settings.py' and 'remote_settings.py' which inherits 'extra_settings.py'.


django_cloud_deploy/skeleton/templates/settings_template/extra_settings.py-tpl
```diff
@@ -0,0 +1,48 @@
...
+
+from .base_settings import *
+
+
+if 'INSTALLED_APPS' in locals():
+    INSTALLED_APPS += [
+        'cloud_admin.apps.CloudAdminConfig',
+    ]
+
+if 'TEMPLATES' in locals():
```
enriquejosepadilla: what if they dont have templates/staticfiles_dirs ?

django_cloud_deploy/skeleton/templates/settings_template/extra_settings.py-tpl
```diff
@@ -0,0 +1,53 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from .base_settings import *
+
+
+if 'INSTALLED_APPS' in locals():
```
brianquinlan: But, we don't install our app if INSTALLED_APPS is not defined?

django_cloud_deploy/skeleton/templates/settings_template/extra_settings.py-tpl
```diff
@@ -0,0 +1,53 @@
...
+    INSTALLED_APPS += [
+        'cloud_admin.apps.CloudAdminConfig',
+    ]
+
+if 'TEMPLATES' not in locals():
+    TEMPLATES = []
+
+TEMPLATES += [
+    {
```
brianquinlan: What does this do?

wangxf123456: W created a directory to save the templates used to overwrite the admin app. This will make those templates available.

django_cloud_deploy/skeleton/templates/settings_template/extra_settings.py-tpl
wangxf123456: I modified the file to make it handle undefined 'INSTALLED_APPS'.

django_cloud_deploy/skeleton/templates/settings_template/extra_settings.py-tpl
brianquinlan: Could you add that as a comment? Maybe:
```suggestion
# Add the templates required by CloudAdminSite.
```

Issue #107, opened 2018-11-29
After installing gcloud, requirement still shows as missing.


enriquejosepadilla:
My hypothesis is because it modified the bash and the user needs to open a new terminal. We should have better documentation.

enriquejosepadilla:
Tell users once gcloud has been installed to open a new terminal to run the commands.

enriquejosepadilla:
100% is due to needing a new terminal.

wangxf123456:
On https://cloud.google.com/sdk/docs/downloads-interactive#linux, where we direct users to install gcloud, it did ask users to restart their terminal.

enriquejosepadilla:
I think I will make it bold and stand out they need to restart their terminal.

brianquinlan:
Check if gcloud is installed but not on path and, if so, then prompt for shell restart.

brianquinlan:
Or just use it if it is in the usual location...or try to find it from the config file.

PR #116, opened 2018-12-03
Check if folders from gcloud installation exists, and remind user to open a new terminal

enriquejosepadilla:
#107

### DIFF
--- a/django_cloud_deploy/tool_requirements.py
+++ b/django_cloud_deploy/tool_requirements.py
@@ -129,12 +129,25 @@ class Gcloud(Requirement):
         Raises:
             MissingRequirementError: If the requirement is not found.
         """
-        if not shutil.which('gcloud'):
-            download_link = (
-                'https://cloud.google.com/sdk/docs/downloads-interactive')
-            msg = ('Please download Google Cloud SDK '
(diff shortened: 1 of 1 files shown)


django_cloud_deploy/tool_requirements.py
```diff
@@ -129,13 +129,26 @@ def check(cls):
...
-            download_link = (
-                'https://cloud.google.com/sdk/docs/downloads-interactive')
-            msg = ('Please download Google Cloud SDK '
-                   'from {}'.format(download_link))
+        if shutil.which('gcloud'):
+            return None
+
+        # Default paths
+        gcloud_config_path = os.path.expanduser('~/.config/gcloud')
+        gcloud_sdk_path = os.path.expanduser('~/.google-cloud-sdk/gcloud')
```
wangxf123456: It should be '~/google-cloud-sdk/gcloud'. (No dot before google-cloud-sdk).

Also, if gcloud is installed with apt-get, then the path of gcloud is '/usr/bin/gcloud'. I am not sure whether this matters.

enriquejosepadilla: ty, good catch!

Issue #112, opened 2018-11-29
Docker Requirement confuses users.


enriquejosepadilla:
I'm actually not sure how docker is installed for users but it gives an error when trying to use it. Double check.

enriquejosepadilla:
https://docs.docker.com/docker-for-mac/install/#install-and-run-docker-for-mac maybe point them to that page.

wangxf123456:
I think installing docker on Mac is completely different from Linux. 

Also, I saw one user get confused when running "sudo usermod -a -G docker $USER". This command does not exist on mac and actually this step is unnecessary on mac.

brianquinlan:
It seems like it isn't the requirement that is confusing but how to actually do the installation?

brianquinlan:
Maybe the "sudo usermod..." thing is only for Googlers and we should remove it.

brianquinlan:
Also, if the user is using the App Engine flag, maybe we shouldn't require docker?

enriquejosepadilla:
(Testing if replying creates a new comment)

Yeah thats another thing, the requirements are different, the user only
needs gcloud.

PR #117, opened 2018-12-03
Remove confusing msg for docker requirement

enriquejosepadilla:
#112

### DIFF
--- a/django_cloud_deploy/tool_requirements.py
+++ b/django_cloud_deploy/tool_requirements.py
@@ -18,6 +18,7 @@ import getpass
 import os
 import shutil
 import stat
+import subprocess
 import sys
 
 import pexpect
@@ -177,9 +178,12 @@ class Docker(Requirement):
                 args = ['group', 'docker']
                 p = pexpect.spawn('getent', args)
                 user = getpass.getuser()
-                if p.expect(user):
-                    return
-        except (pexpect.exceptions.TIMEOUT, pexpect.exceptions.EOF) as e:
+                p.expect(user)
+                command = ['docker', 'image', 'ls']
+                subprocess.check_call(command,
+                                      stdout=subprocess.DEVNULL,
+                                      stderr=subprocess.DEVNULL)
+        except (pexpect.exceptions.TIMEOUT, pexpect.exceptions.EOF):
             link = 'https://docs.docker.com/install/linux/linux-postinstall/'  # noqa
             msg = ('Docker is installed but we are unable to use it.\n'
                    'Please follow {} for more information.\n'
@@ -188,7 +192,12 @@ class Docker(Requirement):
                    'sudo usermod -a -G docker $USER\n'
                    'IMPORTANT: Log out/Log back in after'.format(link))
             raise MissingRequirementError(cls.NAME, msg)
-
+        except subprocess.CalledProcessError:
+            msg = ('You have recently added yourself to the docker '
+                   'group. Please log out/log back in.')
+            raise MissingRequirementError(cls.NAME, msg)
+        finally:
+            p.close()
 
 
 class CloudSqlProxy(Requirement):
--- a/django_cloud_deploy/tool_requirements.py
+++ b/django_cloud_deploy/tool_requirements.py
@@ -172,14 +172,8 @@ class Docker(Requirement):
             subprocess.check_call(command, stdout=subprocess.DEVNULL,
                                   stderr=subprocess.DEVNULL)
         except subprocess.CalledProcessError:
-            # TODO: Handle for multiple OS's
-            # TODO: Check output for error message given when user has
-            # not ran the command
-            msg = ('Docker is installed but not correctly set up.'
-                   'Use the following command to fix it: \n'
-                   '$ sudo groupadd docker\n'
-                   '$ sudo usermod -a -G docker $USER\n'
-                   'Then log out/log back in')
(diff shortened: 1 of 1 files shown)


enriquejosepadilla:
I will test with my mac tommorow, I think testing with linux might be difficult since we are part of the corp and it needs a different installation? 

@brianquinlan @wangxf123456

enriquejosepadilla:
@wangxf123456 being a boss and testing it on vmware for linux. We are very confifdent the command is not needed for mac, but I'll do the mac testing tommorow and confirm.

enriquejosepadilla:
commands are required for linux

django_cloud_deploy/tool_requirements.py
```diff
@@ -157,14 +157,16 @@ def check(cls):
...
-                   'Use the following command to fix it: \n'
-                   '$ sudo groupadd docker\n'
-                   '$ sudo usermod -a -G docker $USER\n'
-                   'Then log out/log back in')
+            if sys.platform.startswith('linux'):
+                msg = ('Docker is installed but not correctly set up.'
+                       'Use the following command to fix it: \n'
+                       '$ sudo groupadd docker\n'
+                       '$ sudo usermod -a -G docker $USER\n'
+                       'Then log out/log back in')
```
brianquinlan: I don't think that you should have to do that. You could also add a link to here:
https://docs.docker.com/install/linux/linux-postinstall/

django_cloud_deploy/tool_requirements.py
```diff
@@ -157,14 +157,16 @@ def check(cls):
...
         except subprocess.CalledProcessError:
-            # TODO: Handle for multiple OS's
-            # TODO: Check output for error message given when user has
-            # not ran the command
-            msg = ('Docker is installed but not correctly set up.'
-                   'Use the following command to fix it: \n'
-                   '$ sudo groupadd docker\n'
-                   '$ sudo usermod -a -G docker $USER\n'
-                   'Then log out/log back in')
+            if sys.platform.startswith('linux'):
```
brianquinlan: You could also do `getent group docker` and see if the user's name is in it. Not sure if that is overkill or not.

django_cloud_deploy/tool_requirements.py
```diff
@@ -157,14 +157,16 @@ def check(cls):
...
+                msg = ('Docker is installed but not correctly set up.'
```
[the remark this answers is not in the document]
enriquejosepadilla: Nope.

django_cloud_deploy/tool_requirements.py
```diff
@@ -157,14 +157,16 @@ def check(cls):
...
+                       'Use the following command to fix it: \n'
+                       '$ sudo groupadd docker\n'
+                       '$ sudo usermod -a -G docker $USER\n'
+                       'Then log out/log back in')
+            else:
+                msg = ('Docker is installed but was unable to run the command.'
+                       'Make sure docker is running, if it is open '
```
brianquinlan: Maybe "Make sure that docker is running and on the path. If you recently installed it, you may have to <whatever you need to do to ensure that the path gets updated>."

django_cloud_deploy/tool_requirements.py
enriquejosepadilla: Is there any gain from doing that? Thats a subprocess command as well right?

enriquejosepadilla:
I changed CLA since its my account from my mac, i thought i had fixed it though. Will check in a test PR

django_cloud_deploy/tool_requirements.py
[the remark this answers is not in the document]
brianquinlan: Then you probably want a space after the period.

django_cloud_deploy/tool_requirements.py
brianquinlan: You are recommending that the user do because you got an error from docker. But you don't actually know that it is related to group membership. `getent` would allow you to check.

django_cloud_deploy/tool_requirements.py
```diff
@@ -177,19 +184,30 @@ def check(cls):
             raise MissingRequirementError(cls.NAME, msg)
 
         try:
-            command = ['docker', 'image', 'ls']
-            subprocess.check_call(command, stdout=subprocess.DEVNULL,
-                                  stderr=subprocess.DEVNULL)
+            if sys.platform.startswith('linux'):
```
brianquinlan: I don't think that you doc comment is true - you actually check that docker works on Linux . I would structure the code like this:

```python

def _correct_docker_group_setup(self) -> bool:
  user = getpass.getuser()
  p = pexpect.spawn('getent', args)
  try:
    p.expect(user)
  except pexpect.exceptions.EOF:
    return False
  else:
    return True
    
try:
  subprocess.check_call(['docker', 'image', 'ls'])
subprocess.CalledProcessError:
  if sys.platform.statswith('linux') and not self._correct_docker_group_setup():
    msg = ('Docker is installed but you "{user}" are not in the "docker" group, which is
               'required when running docker outside without superuser permissions.'
               'You can run the following commands to fix it: \n'
               '...'
               'For more information, refer to: {link}')
      raise MissingRequirementError(cls.NAME, msg)
    else:
      msg = 'Docker is installed by is not usable. Please follow the installation instructions'
                'at {link}'          
```

PR #124, opened 2018-12-04
Update GETTING-STARTED.md

brianquinlan:


GETTING-STARTED.md
```diff
@@ -1,4 +1,4 @@
-# Getting Started
+# Getting Started
```
rickeylev: 1. I'd omit the leading `$` from the command excerpts. unlike g3doc, github markdown doesn't make the leading `$` non-selectable, so it makes copy/paste harder.
2. Provide copy/paste commands to install the prereqs, if possible
3. I'd suggest -U to the pip install command. That'll ensure people get an up-to-date version of your tool

enriquejosepadilla:
@brianquinlan Should this be closed?

GETTING-STARTED.md
brianquinlan: Yes, I fixed this in my latest PR.

PR #126, opened 2018-12-05
Add a sanitizer for GCP resource names.

### LINKED ISSUE #108: Project name should adhere to projectid validation
User study had an error for instance name: django-on-gke4:Django_on_GKE4-instance 

Most likely it should be lower case and no underscores. We should validate this so it doesn't cause issues.

wangxf123456:
Fixes #108


django_cloud_deploy/workflow/__init__.py
```diff
@@ -306,6 +311,18 @@ def update_project(self,
         if open_browser:
             webbrowser.open(app_url)
 
+    @staticmethod
+    def _sanitize_name(name: str) -> str:
```
enriquejosepadilla: is this enough? What if the user uses non alphanumeric values? "#:)Break/Stuff "

wangxf123456: This function is only used to sanitize django_project_name. When prompting users for this, we use django_project_name.isidentifier() to validate it. See https://github.com/GoogleCloudPlatform/django-cloud-deploy/blob/master/django_cloud_deploy/cli/prompt.py#L192. This will make sure django_project_name only contains lowercase or uppercase characters, numbers or underscores.

Issue #102, opened 2018-11-29
Requirements are unclear if unmet, be very clear to user if a requirement is missing.


enriquejosepadilla:

enriquejosepadilla:
1. There was an error before telling the user they didnt have gcloud
2. If they dont have gcloud, that means cloud sql will never be on their path.

Issue #67, opened 2018-11-12
tool_requirements.py asks user to install cloud_sql_proxy manually after it is installed by our tool


wangxf123456:
Output of the tool:

$ django-cloud-deploy new
Cloud Sql Proxy is required, download? [Y/n]:y
Please install the following requirements:
* Gcloud SDK: Please download Google Cloud SDK from https://cloud.google.com/sdk/docs/downloads-interactive
* Docker: Please download Docker from https://store.docker.com/
* Cloud Sql Proxy: Please download Cloud SQL Proxy from https://cloud.google.com/sql/docs/mysql/sql-proxy

enriquejosepadilla:
It also asked participants to redownload dependencies they had downloaded in previous steps.

PR #127, opened 2018-12-05
Clean Cloud Sql Proxy Requirement

enriquejosepadilla:
Improve cloud sql proxy message to users as well as use gcloud to download.


enriquejosepadilla:
> ```shell
> googleapiclient.errors.HttpError: <HttpError 503 when requesting https://www.googleapis.com/upload/storage/v1/b/bucket-20181205211450-s8mp/o?uploadType=multipart&alt=json returned "Backend Error">
> ```
> 
> I am afraid there is something wrong with the google storage api. Maybe we should wait for a while and run the test again.

:( Ok

enriquejosepadilla:
#67 #102

PR #128, opened 2018-12-05
Add progress bar context manager to console io.

wangxf123456:


django_cloud_deploy/cli/io.py
```diff
@@ -14,10 +14,75 @@
...
+    def start(self):
+        self._thread.start()
+
+    def finish(self):
+        self._finish()
+
+    def _run(self):
+        """The function to update progress bar."""
+        for _ in range(self._expected_time * 2):
+            self._bar_lock.acquire()
```
brianquinlan: I think that you can use the lock with a `with` statement

django_cloud_deploy/cli/io.py
```diff
@@ -14,10 +14,75 @@
...
+            self._bar_lock.release()
+            time.sleep(0.5)
+
+    def _finish(self):
+        """Make progress bar go to the end.
+
+        This is useful when the task takes shorter than the expect time to
+        finish.
+        """
+        self._bar_lock.acquire()
```
brianquinlan: Use `with`

django_cloud_deploy/cli/io.py
```diff
@@ -14,10 +14,75 @@
...
+        # The bar will tick every 0.5 seconds.
+        self._bar = bar.ChargingBar(message, max=expect_time * 2)
+        self._thread = threading.Thread(target=self._run)
+        self._bar_lock = threading.Lock()
+
+    def start(self):
+        self._thread.start()
+
+    def finish(self):
+        self._finish()
```
brianquinlan: Why the level of indirection?

django_cloud_deploy/cli/io.py
```diff
@@ -14,10 +14,75 @@
...
+import time
+
+from progress import bar
+
+
+class _ProgressBar(object):
+    """A progress bar showing status of a task.
+
+    Output of the progress bar will be like the following:
+        Processing █████████████████∙∙∙∙∙∙∙∙∙∙∙∙∙∙∙ 55%
```
brianquinlan: You aren't going to include a time estimate? (at least optionally)

django_cloud_deploy/cli/io.py
```diff
@@ -80,9 +145,18 @@ def ask(self, prompt=None):
         return input(self._replace_html_tags(prompt, sys.stdout.fileno()))
 
     def getpass(self, prompt=None):
-        "Prompt the user for a password and return the result."
+        """Prompt the user for a password and return the result."""
         return getpass.getpass(prompt)
 
+    @contextlib.contextmanager
+    def progressbar(self, expected_time: int, message: str):
```
brianquinlan: Could you write a doc string? Also, are you going to check to see if you are writing to a tty in a follow-up CL?

django_cloud_deploy/cli/io.py
wangxf123456: Done.

django_cloud_deploy/cli/io.py
wangxf123456: Done.

django_cloud_deploy/cli/io.py
wangxf123456: Removed one of the methods.

django_cloud_deploy/cli/io.py
wangxf123456: I added time estimate.

django_cloud_deploy/cli/io.py
wangxf123456: Done.

django_cloud_deploy/cli/io.py
```diff
@@ -80,9 +151,39 @@ def ask(self, prompt=None):
...
+            expect_time: How long the progress bar is going to run.
+                (In seconds).
+            message: A prefix of the progress bar showing what it is about.
+
+        Yields:
+            None
+        """
+
+        if os.isatty(sys.stdout.fileno()) and os.name == 'posix':
+            progress_bar = _ProgressBar(expect_time, message, posix=True)
```
brianquinlan: So the idea behind isatty is to determine whether this is an interactive console or not.

Does the progress bar work when writing to a text file e.g.
`$ django-cloud-deploy $(ARGS) >myfile.txt`

?

wangxf123456: Done. Only print estimate time to finish if not in an interactive console.

Issue #130, opened 2018-12-06
Docker isnt needed for GAE


enriquejosepadilla:

wangxf123456:
Fixed by #134

PR #134, opened 2018-12-10
Remove docker dependency on GAE

enriquejosepadilla:


enriquejosepadilla:
fixes #130

django_cloud_deploy/tool_requirements.py
```diff
@@ -263,25 +263,33 @@ def handle(cls, console: io.IO):
...
+    ]
+}
+
+
+def check_and_handle_requirements(console: io.IO, backend: str):
     """Checks that requirements are installed. Attempts to install missing ones.
 
     Args:
         console: IO, class that handles the input/output with the user.
+        backend: str, Defines which platform on Determines what requirements are
```
wangxf123456: Why "Determines" is capitalized?

Also, maybe we do not need to put type information here. It shows as type annotation for parameters.

Issue #133, opened 2018-12-08
E2E test about update command fails sometimes


wangxf123456:
```bash
../.nox/e2e_test-python_version-3-5/lib/python3.5/site-packages/absl/testing/absltest.py:1172: in fail
    return super(TestCase, self).fail(self._formatMessage(prefix, msg))
E   AssertionError: 'Hello1 from the Cloud!' not found in '<html xmlns="http://www.w3.org/1999/xhtml"><head></head><body></body></html>'
```

PR #135, opened 2018-12-10
Fix e2e test

wangxf123456:
The e2e test sometimes fails with the following error message:
```.pytb
AssertionError: 'Hello1 from the Cloud!' not found in '<html xmlns="http://www.w3.org/1999/xhtml"><head></head><body></body></html>'
```

Looks like this is because there is a bug related to selenium. Some people reported the same bug on stackoverflow.

- https://stackoverflow.com/questions/47715071/chrome-headless-in-unix-returns-empty-page-source
- https://stackoverflow.com/questions/23581291/python-selenium-with-phantomjs-empty-page-source

So I decided to not using selenium to get the page source.


wangxf123456:
Fixes #133

PR #148, opened 2018-12-13
Dump traceback to a temporary file.

wangxf123456:
The bug report will be dumped at "/tmp/django-deploy-bug-report-[random suffix]".

Also removed "try again" when internal errors happened because most of the time "try again" does not solve the problem.


django_cloud_deploy/crash_handling/__init__.py
```diff
@@ -49,12 +50,22 @@ def handle_crash(err: Exception, command: str,
...
+        prefix='django-deploy-bug-report-')
     console.tell(
-        ('Your "{}" failed due to an internal error - please try again.'
+        ('Your "{}" failed due to an internal error.'
          '\n\n'
          'You can report this error by filing a bug on Github. If you agree,\n'
          'a browser window will open and an Github issue will be\n'
-         'pre-populated with the details of this crash.\n').format(command))
+         'pre-populated with the details of this crash.\n'
+         'See bug report at "{}" for more information.').format(
```
brianquinlan:
```suggestion
         'For more details, see: {}').format(
```

django_cloud_deploy/crash_handling/__init__.py
```diff
@@ -49,12 +50,22 @@ def handle_crash(err: Exception, command: str,
...
+             command, log_file_path))
+    issue_content = _create_issue_body(command)
```
brianquinlan: I'd write the file before telling the user about it.

django_cloud_deploy/crash_handling/__init__.py
wangxf123456: Done.

django_cloud_deploy/crash_handling/__init__.py
wangxf123456: Done.

PR #149, opened 2018-12-13
Clean up prompt.py.

wangxf123456:
Added more user input validations and removed unused code.


django_cloud_deploy/cli/prompt.py
```diff
@@ -662,7 +633,14 @@ def prompt(cls,
         """
         console_prompt = cls._get_prompt(arguments)
         console.tell(('{} {}').format(step_prompt, console_prompt))
-        return console.getpass('Password: ')
+        while True:
+            password = console.getpass('Password: ')
```
brianquinlan: If this is the user's postgres password then I'd write that. Like `Postgres password: `

PR #154, opened 2018-12-14
Simplify automatic cloud_sql_proxy installation.

brianquinlan:


django_cloud_deploy/tool_requirements.py
```diff
@@ -256,34 +262,24 @@ def handle(cls, console: io.IO):
...
         """
         if shutil.which('gcloud') is None:
-            msg = "Gcloud is needed to install Cloud Sql Proxy"
+            msg = "gcloud is needed to install Cloud SQL Proxy"
             raise UnableToAutomaticallyInstallError(cls.NAME, msg)
 
         while True:
-            answer = console.ask('Cloud Sql Proxy is required by Django Cloud '
-                                 'Deploy. Would you like us to install '
+            answer = console.ask('Cloud SQL Proxy is required by Django '
```
enriquejosepadilla: do we tell them default is yes?

brianquinlan: Isn't that why 'Y' is uppercase?

PR #161, opened 2018-12-19
[WIP] Fix flaky test.

wangxf123456:


django_cloud_deploy/tests/e2e/gke_deploy_test.py
```diff
@@ -16,10 +16,12 @@
 import os
 import shutil
 import tempfile
+import time
```
enriquejosepadilla: where do you use this import?

wangxf123456: This is not used. I removed it.

PR #172, opened 2019-01-08
Enable pytype for django-cloud-deploy.

rchen152:
The unreadable files issue (https://github.com/google/importlab/issues/37) that was causing pytype to fail before even starting type-checking has been fixed, so I decided to take a stab at enabling pytype. Let me know what you think.

Note that pytype found a couple of what I think are real type errors (the last two bullet points). Since I'm not familiar with the code, I silenced them, but I'm happy to make real fixes if given some guidance on how to do so.

* Adds the pytype output directory to .gitignore. I set a custom output directory because the default, `pytype_output`, will likely be changed at some point (https://github.com/google/pytype/issues/202).
* Silences a false positive in `django_cloud_deploy/crash_handling/__init__.py`.
* Changes the nox file so that pytype runs as `nox -f django_cloud_deploy/nox.py -s type_check`. While I was at it, I also fixed the `lint` session.
* Moves pytype options to setup.cfg so that pytype can easily be run directly without going through nox. Also disables an error class, `invalid-typevar`, that was causing two false positives.
* Silences a real finding in `django_cloud_deploy/skeleton/source_generator.py`. `@abc.abstractmethod` is being used in a class that does not have `abc.ABCMeta` as its metaclass, so the abstractness of the methods isn't being enforced. The fix isn't as simple as adding the metaclass, since supposedly abstract classes are instantiated later in the file.
* Fixes an incorrect annotation in `django_cloud_deploy/workflow/__init__.py` and silences two real findings. (A parameter marked as `Optional` is being used without a `None` check.)


enriquejosepadilla:
This is a lovely surprise :+1:

django_cloud_deploy/workflow/__init__.py
```diff
@@ -417,16 +417,19 @@ def _generate_base_secrets(database_username: str,
 
     @staticmethod
     def _load_secret_names(required_service_accounts: Optional[
```
brianquinlan: I think that you  can remove the `Optional` - I looked at the call site and it can never be None.

brianquinlan: Then you can remove the two disables.

enriquejosepadilla:
@rchen152 Im not sure why the lint/integration/unit checks weren't run, but hopefully when you resolve the comments and commit again they get run. If not ill look into it.

django_cloud_deploy/workflow/__init__.py
rchen152: Don!

django_cloud_deploy/workflow/__init__.py
```diff
@@ -416,8 +416,9 @@ def _generate_base_secrets(database_username: str,
         return {'username': database_username, 'password': database_password}
 
     @staticmethod
-    def _load_secret_names(required_service_accounts: Optional[
-            Dict[str, List[Dict[str, Any]]]] = None) -> (List[str], List[str]):
+    def _load_secret_names(
+            required_service_accounts: Dict[str, List[Dict[str, Any]]]
+    ) -> Tuple[List[str], List[str]]:
```
brianquinlan: Was lint happy with this?

rchen152: Should be, this is the formatting I got from doing `yapf -i`.

enriquejosepadilla: when i need to lint fix i run yapf -r -i cli cloudlib workflow tests

enriquejosepadilla: I believe it was optional because of testing.

rchen152: I ran `yapf -r -i cli cloudlib workflow tests` and there was no diff. How do I see why the Kokoro builds failed? When I click on "Details", I get a page not found error.

brianquinlan:
It looks like the CI might be broken. I'll merge this and we can fix any problems that turn up later. Thanks so much Rebecca!

Issue #169, opened 2018-12-21
Refactor Prompts


enriquejosepadilla:
The current state of prompts is starting to limit the flexibility of changes. 

They are also IMO getting a bit complex.

PR #173, opened 2019-01-09
Prompt Refactor Base

enriquejosepadilla:
The idea is that prompt managers use these classes to prompt the user for one parameter.

The Prompt classes (these) are responsible only of asking the user for input. Some of them take a validation function that might make api calls, but the prompt doesn't know that. The goal was to make the prompts not responsible of business logic or contain api calls.


enriquejosepadilla:
Part of #169

django_cloud_deploy/cli/prompt_new.py
```diff
@@ -0,0 +1,204 @@
...
+            ValueError: if the input string is not valid.
+        """
+        pass
+
+
+class AskPrompt(Prompt):
+    """Used to ask for a single string value.
+
+    Attributes:
+        question: Question shown to the user on the console.
```
brianquinlan: Are these meant to be public?

django_cloud_deploy/cli/prompt_new.py
```diff
@@ -0,0 +1,204 @@
...
+        default: Default value if user provides no value. (Presses enter)
+    """
+
+    def __init__(self,
+                 question: str,
+                 console: io.IO,
+                 validate: Optional[Callable[[str], bool]] = None,
+                 default: Optional[str] = None):
+        self.question = question
+        self.validate = validate or (lambda s: True)
```
brianquinlan: So you are overriding a public method with an instance variable. Maybe:
```
def __init__(self, ...):
  self._validate_fn = ...

def validate(self, s):
  self._validate_fn(s)
```

django_cloud_deploy/cli/prompt_new.py
```diff
@@ -0,0 +1,204 @@
...
+        except ValueError as e:
+            self.console.error(e)
+            return self.prompt()
+
+
+class MultipleChoicePrompt(Prompt):
+    """Used to prompt user to choose from a list of values.
+
+    Attributes:
+        question: Question shown to the user on the console.
```
brianquinlan: Are these meant to be public?

django_cloud_deploy/cli/prompt_new.py
```diff
@@ -0,0 +1,204 @@
...
+        """
+        options_formatted = [
+            '{}. {}'.format(str(i), opt)
+            for i, opt in enumerate(self.options, 1)
+        ]
+        options = '\n'.join(options_formatted)
+        answer = self.console.ask(self.question.format(options))
+        try:
+            self.validate(answer)
+            return answer
```
brianquinlan: Move this out of the try block

django_cloud_deploy/cli/prompt_new.py
```diff
@@ -0,0 +1,204 @@
...
+
+    Attributes:
+        question: Question shown to the user on the console.
+        options: Possible values user should choose from.
+        console: Object to use for user I/O.
+        default: Default value if user provides no value. (Presses enter)
+    """
+
+    def __init__(self,
+                 question: str,
```
brianquinlan: You should document the required format for `question`

django_cloud_deploy/cli/prompt_new.py
```diff
@@ -0,0 +1,204 @@
...
+
+        Returns:
+            The choice entered by the user.
+        """
+        options_formatted = [
+            '{}. {}'.format(str(i), opt)
+            for i, opt in enumerate(self.options, 1)
+        ]
+        options = '\n'.join(options_formatted)
+        answer = self.console.ask(self.question.format(options))
```
brianquinlan: Do we want to show all of the options when the user enters a bad value

brianquinlan: Also, your approach for handling defaults is different for BinaryPrompt.

django_cloud_deploy/cli/prompt_new.py
```diff
@@ -0,0 +1,204 @@
...
+        """
+        answer = self.console.ask(self.question)
+        if self.default and answer is '':
+            answer = self.default
+        if self.validate(answer):
+            return answer
+        else:
+            return self.prompt()
+
+    def validate(self, s: str):
```
brianquinlan: Will this work when validating flags?

django_cloud_deploy/cli/prompt_new.py
```diff
@@ -0,0 +1,204 @@
...
+        self.default = default
+
+    def prompt(self):
+        """Prompt the user to choose from a yes or no question.
+
+        Returns:
+            The choice entered by the user.
+        """
+        answer = self.console.ask(self.question)
+        if self.default and answer is '':
```
brianquinlan:
```suggestion
        if self.default and answer == '':
```

django_cloud_deploy/cli/prompt_new.py
```diff
@@ -0,0 +1,204 @@
...
+        self.console = console
+
+    def prompt(self) -> str:
+        """Prompt the user to enter a password.
+
+        Returns:
+            The value entered by the user.
+        """
+        while True:
+            password1 = self.console.getpass('Password: ')
```
brianquinlan: Don't we use different prompts for database vs admin passwords?

django_cloud_deploy/cli/prompt_manager.py
```diff
@@ -0,0 +1,673 @@
+import abc
```
brianquinlan: Do you have the usage of `prompt_manager` done? Could I see a usage example?

django_cloud_deploy/cli/prompt_manager.py
```diff
@@ -0,0 +1,673 @@
...
+    def validate(self, s: str) -> bool:
+        """Validates that a string is a valid project name.
+
+        Args:
+            s: The string to validate.
+
+        Raises:
+            ValueError: if the input string is not valid.
+        """
+        if not (4 <= len(s) <= 30):
```
brianquinlan: I think that there is more to it than that.

django_cloud_deploy/cli/prompt_manager.py
```diff
@@ -0,0 +1,673 @@
...
+                 'must be between 4 and 30 characters').format(s))
+
+        if self._is_new_project():
+            return True
+
+        project_id = self.args.get('project_id', None)
+        if project_id is None:
+            raise ValueError('Project Id must be set')
+
+        project_name = self.project_client.get_project(project_id)['name']
```
brianquinlan: So we force the user to specify the same name? Why would we want that? Maybe it should be an error that the flag is set when using an existing project.

django_cloud_deploy/cli/prompt_manager.py
```diff
@@ -0,0 +1,673 @@
...
+                               project.ProjectClient.from_credentials(creds))
+
+    def validate(self, s: str) -> bool:
+        """Validates that a string is a valid project name.
+
+        Args:
+            s: The string to validate.
+
+        Raises:
+            ValueError: if the input string is not valid.
```
brianquinlan: I think that this returns a bool and doesn't raise. It is slightly confusing that you use two different approaches for methods with the same name, arguments, etc.

django_cloud_deploy/cli/prompt_manager.py
```diff
@@ -0,0 +1,673 @@
...
+        msg = '\n'.join([msg_base, msg_default])
+        return prompt_new.AskPrompt(
+            msg, self.console, self.validate, default=default_answer)
+
+    def _is_new_project(self):
+        must_exist = workflow.ProjectCreationMode.MUST_EXIST
+        return self.args.get('project_creation_mode', None) != must_exist
+
+    def _handle_existing_project(self):
+        assert 'project_id' in self.args, 'project_id must be set'
```
brianquinlan: This can't be an assert, right? Because the user could really pass

 --project_creation_mode=MUST_EXIST but not specify a project_id.

django_cloud_deploy/cli/prompt_manager.py
enriquejosepadilla: Will add to next commit after fixes.

django_cloud_deploy/cli/prompt_manager.py
```diff
@@ -0,0 +1,673 @@
...
+        return self.args.get('credentials', None)
+
+    def _is_valid_passed_arg(self) -> bool:
+        value = self.args.get(self.PARAMETER, None)
+        if value is None:
+            return False
+
+        try:
+            if self.validate(value):
+                msg = '{} {}: {}'.format(self.step, self.PARAMETER, value)
```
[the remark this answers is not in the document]
enriquejosepadilla: and clever? jaja It mostly solves the problem of having N steps be final instead of a calculated values.

django_cloud_deploy/cli/prompt_manager.py
```diff
@@ -0,0 +1,673 @@
...
+                self.console.tell(msg)
+                return True
```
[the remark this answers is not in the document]
enriquejosepadilla: agreed

django_cloud_deploy/cli/prompt_manager.py
enriquejosepadilla: Im not sure what you mean by this. It does raise. I do think that the bools are a bit superfluous.

django_cloud_deploy/cli/prompt_manager.py
enriquejosepadilla: I believe I copied our previous code. But dont mind adding extra logic.

django_cloud_deploy/cli/prompt_manager.py
enriquejosepadilla: I agree its wierd. Ill think we can just ignore that they gave it to us and use the one the api call gives us.

django_cloud_deploy/cli/prompt_manager.py
enriquejosepadilla: Project creation mode is hard coded to create unless they pass in a project-id.

enriquejosepadilla: In other words, its not a flag available to the user.

django_cloud_deploy/cli/prompt_manager.py
```diff
@@ -0,0 +1,673 @@
...
+
+    PARAMETER = 'billing_account_name'
+
+    def __init__(self,
+                 console: io.IO,
+                 step: str,
+                 args,
+                 billing_client: billing.BillingClient = None):
+        super().__init__(console, step, args)
+        creds = self.get_credentials()
```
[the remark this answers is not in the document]
enriquejosepadilla: actually thatll help me with testing. ty

django_cloud_deploy/cli/prompt_manager.py
enriquejosepadilla: I think this would be more of a safeguard for ourselves as there is not reason it should reach this branch if project_id is not already set by this point. (as well as project_id being prompted before project_name)

django_cloud_deploy/cli/prompt_manager.py
```diff
@@ -0,0 +1,689 @@
...
+            self.args['project_creation_mode'] = (
+                workflow.ProjectCreationMode.MUST_EXIST)
+
+    def prompt(self):
+        self._setup()
+
+        total_steps = len(self.PROMPT_ORDER)
+        for i, prompt in enumerate(self.PROMPT_ORDER, 1):
+            step = '<b>[{}/{}]</b>'.format(i, total_steps)
+            self.required_parameters_to_prompt[prompt](self.console, step,
```
brianquinlan: So args is mutated through a shared dictionary. Yuck.

enriquejosepadilla: probably will call it retrieve_parameters(..., args) -> args. Does that sound good?

django_cloud_deploy/cli/prompt_manager.py
```diff
@@ -0,0 +1,689 @@
...
+
+    def prompt(self) -> None:
+        """Handles the business logic and calls the prompts."""
+        pass
+
+    def validate(self, val: str):
+        """Checks if the string is valid. Throws a ValueError if not valid."""
+        return
+
+    def get_credentials(self):
```
gpshead: Consider an `@property` in Python rather than a getter method.  if there is a need for this to be overridden, `@property` for `.credentials` that just calls a `_get_credentials` method that intended to be overidden in subclasses.

enriquejosepadilla:
@wangxf123456 somewhere in the test one of my prompts must be asking for some extra information but im not sure how to debug it to figure out which one it is. I can run it locally but my print statements arent working, any suggestions of best way to check what needs to be fixed. I know its between ' 
        'project_id',
        'project_name',
        'billing_account_name',
        'database_password',
        'django_directory_path',

but not sure which one :/

enriquejosepadilla:
Will need to transfer over the unit tests, as this is code that only being by integration.

django_cloud_deploy/cli/prompt_manager.py
```diff
@@ -0,0 +1,875 @@
...
+    allowed_characters = frozenset(string.ascii_letters + string.digits +
+                                   string.punctuation)
+    if frozenset(s).issuperset(allowed_characters):
+        raise ValueError('Invalid character in password: '
+                         'use letters, numbers and punctuation')
+
+    return
+
+
+class TemplatePrompt(object):
```
brianquinlan: Should there be a `Prompt` that is an abc that this (which is also an abc) is a subclass of?

django_cloud_deploy/cli/prompt.py
```diff
@@ -32,260 +17,401 @@
...
-        Args:
-            console: Object to use for user I/O.
-            step_prompt: A prefix showing the current step number e.g. "[1/3]".
-            arguments: The arguments that have already been collected from the
-                user e.g. {"project_id", "project-123"}
-            credentials: The OAuth2 Credentials object to use for api calls
-                during prompt.
+def _ask_prompt(question: str,
+                console: io.IO,
+                validate: Optional[Callable[[str], None]] = None,
```
wangxf123456: Maybe just "validate: Optional[Callable[str]]"?

django_cloud_deploy/cli/update.py
```diff
@@ -56,53 +55,14 @@ def main(args: argparse.Namespace, console: io.IO = io.ConsoleIO()):
...
-        parameter_and_prompt = sorted(
-            remaining_parameters_to_prompt.items(),
-            key=lambda i: prompt_order.index(i[0]))
-
-        for step, (parameter_name, prompter) in enumerate(parameter_and_prompt):
-            step = '<b>[{}/{}]</b>'.format(step + 1, num_steps)
-            actual_parameters[parameter_name] = prompter.prompt(
-                console, step, actual_parameters)
+    root_prompt = prompt.RootPrompt()
+    actual_parameters = root_prompt.prompt(prompt.Command.UPDATE, console,
```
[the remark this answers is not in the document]
enriquejosepadilla: What do you mean by duplicating arguments?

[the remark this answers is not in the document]
brianquinlan: I mean that `new`, `update` and `cloudify` share arguments but each site must redefine them.

django_cloud_deploy/cli/prompt.py
```diff
@@ -32,260 +32,397 @@
...
-    def _default_name(arguments: Dict[str, Any]):
-        """The default name to use."""
+    if 1 <= int(s) <= (len_options + 1):
+        return
+    else:
+        raise ValueError('Value is not in range')
+
+
+def _binary_prompt(question: str, console: io.IO,
+                   default: Optional[str] = None) -> str:
```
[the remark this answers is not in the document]
enriquejosepadilla: Yes, so is multiple choice prompt.

django_cloud_deploy/cli/prompt.py
```diff
@@ -32,260 +32,397 @@
...
+    Returns:
+        The choice entered by the user.
+    """
+
+    while True:
+        try:
+            answer = console.ask(question)
+            if default and not answer:
+                answer = default
+                _binary_validate(answer)
```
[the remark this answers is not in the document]
enriquejosepadilla: If default is set then there's nothing to validate, the user should handle the default option as an answer.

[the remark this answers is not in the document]
enriquejosepadilla: Yes.

django_cloud_deploy/cli/prompt.py
```diff
@@ -32,260 +32,397 @@
...
-        message = 'Project name found: {}'.format(project_name)
-        console.tell(('{} {}').format(step_prompt, message))
-        return project_name
+        pass
 
-    @staticmethod
-    def validate(s, credentials: Optional[credentials.Credentials] = None):
-        """Validates that a string is a valid project name.
+    @abc.abstractmethod
+    def _is_valid_passed_arg(self, console: io.IO, step: str,
```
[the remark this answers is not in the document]
enriquejosepadilla: It is the function that checks if the passed in argument via the command line is valid. Updated the one liner to reflect that.

django_cloud_deploy/cli/prompt.py
```diff
@@ -32,260 +32,397 @@
...
+            A boolean indicating if the passed in argument is valid.
+        """
+        if value is None:
+            return False
 
+        try:
+            validate(value)
+        except ValueError as e:
+            console.error(e)
+            return False
```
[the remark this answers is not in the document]
enriquejosepadilla: If its invalid it means most likely something will break in the process, so either we quit them or we ask for it again.

django_cloud_deploy/cli/prompt.py
```diff
@@ -32,260 +32,397 @@
...
-    @classmethod
-    def _default_name(cls, arguments: Dict[str, Any]):
-        return 'admin'
+    def __init__(self, project_client: project.ProjectClient):
+        self.project_client = project_client
 
-    @staticmethod
-    def validate(s, credentials: Optional[credentials.Credentials] = None):
-        """Validates that a string is a valid Django superuser login.
+    def _validate(self, project_id: str,
```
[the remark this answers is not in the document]
enriquejosepadilla: Coolios, i was thinking currying which is quite similar, but this works as well :)

django_cloud_deploy/cli/prompt.py
```diff
@@ -32,260 +32,397 @@
...
+
+            if self._is_new_project(project_creation_mode):
+                return
+
+            if project_id is None:
+                raise ValueError('Project Id must be set')
+
+            project_name = self.project_client.get_project(project_id)['name']
+            if project_name != s:
+                raise ValueError('Wrong project name given for project id.')
```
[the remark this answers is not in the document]
enriquejosepadilla: Validate currently doesnt do warning, should it be a new type of error that is meant as warning? 

It is technically wrong if they passed in a wrong project name, if they are unsure or dont care, they dont need to pass in the project name.

django_cloud_deploy/cli/prompt.py
```diff
@@ -32,260 +32,397 @@
...
+        console.tell(message)
+        return project_name
 
-class DjangoSuperuserEmailPrompt(NamePrompt):
-    """Allow the user to enter a Django email address."""
+    def prompt(self, console: io.IO, step: str,
+               args: Dict[str, Any]) -> Dict[str, Any]:
+        new_args = copy.deepcopy(args)
+
+        project_id = args.get('project_id', None)
```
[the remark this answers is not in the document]
enriquejosepadilla: this will throw an error if project id is not set.

django_cloud_deploy/cli/prompt.py
```diff
@@ -32,260 +32,397 @@
...
+        project_name = self.project_client.get_project(project_id)['name']
+        message = '{} {}: {}'.format(step, self.PARAMETER, project_name)
+        console.tell(message)
+        return project_name
 
-class DjangoSuperuserEmailPrompt(NamePrompt):
-    """Allow the user to enter a Django email address."""
+    def prompt(self, console: io.IO, step: str,
+               args: Dict[str, Any]) -> Dict[str, Any]:
+        new_args = copy.deepcopy(args)
```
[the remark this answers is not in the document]
enriquejosepadilla: dict(args) is a shallow copy, currently it would work, but in the future? this techically works now and in the future.

django_cloud_deploy/cli/prompt.py
```diff
@@ -32,260 +32,397 @@
...
+
+        project_id = args.get('project_id', None)
+        project_creation_mode = args.get('project_creation_mode', None)
```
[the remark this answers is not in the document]
enriquejosepadilla: same as above.

django_cloud_deploy/cli/prompt.py
```diff
@@ -32,260 +32,397 @@
...
+        if self._is_valid_passed_arg(
+                console, step, args.get(self.PARAMETER, None),
+                self._validate(project_id, project_creation_mode)):
+            return new_args
+
+        project_creation_mode = args.get('project_creation_mode', None)
```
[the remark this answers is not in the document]
enriquejosepadilla: same as above.

django_cloud_deploy/cli/prompt.py
```diff
@@ -295,239 +432,382 @@ def _generate_default_project_id(project_name=None):
...
+
+        console.tell(
+            ('{} In order to deploy your application, you must allow Django '
+             'Deploy to access your Google account.').format(step))
+        create_new_credentials = True
+        active_account = self.auth_client.get_active_account()
+
+        msg = ('You have logged in with account [{}]. Do you want to '
+               'use it? [Y/n]: ').format(active_account)
+        use_active_credentials = _binary_prompt(msg, console, default='Y')
```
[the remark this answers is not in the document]
enriquejosepadilla: wups

django_cloud_deploy/cli/update.py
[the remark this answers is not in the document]
enriquejosepadilla: Ah, yeah.  make a issue about it?

django_cloud_deploy/cli/prompt.py
```diff
@@ -295,239 +432,382 @@ def _generate_default_project_id(project_name=None):
...
         Args:
             s: The string to validate.
-            credentials: The OAuth2 Credentials object to use for api calls
-                during validation.
 
         Raises:
             ValueError: if the input string is not valid.
         """
+
         if not re.match(r'[a-z][a-z0-9\-]{5,29}', s):
```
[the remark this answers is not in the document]
enriquejosepadilla: Why not ?

django_cloud_deploy/cli/update.py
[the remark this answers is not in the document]
brianquinlan: Yes please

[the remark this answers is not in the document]
enriquejosepadilla: #228

django_cloud_deploy/cli/prompt.py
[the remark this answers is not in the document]
enriquejosepadilla: True, but havent you ever forgotten a string and its because it has wierd things that you forget? If the user uses my_string theyll be reminded its my-string (hopefully)

django_cloud_deploy/cli/prompt.py
```diff
@@ -32,260 +33,430 @@
...
+
+    Typical usage:
+        # User can press enter if user doesn't want anything.
+        choice = _multiple_choice_prompt('Choose an option:\n{}\n',
+                                         ['Chicken', 'Salad', 'Burger'],
+                                         console,
+                                         default=None)
+
+    Returns:
+        The choice made by the user. Will be an index in the options list unless
```
brianquinlan: If there is no default set then shouldn't we force the user to choose (instead of returning `None`)? That's the way other prompts work.

django_cloud_deploy/cli/prompt.py
```diff
@@ -32,260 +33,430 @@
...
 
-        Args:
-            s: The string to validate.
-            credentials: The OAuth2 Credentials object to use for api calls
-                during validation.
+    options_formatted = [
+        '{}. {}'.format(str(i), opt) for i, opt in enumerate(options, 1)
+    ]
+    options = '\n'.join(options_formatted)
+    answer = console.ask(question.format(options))
```
brianquinlan: You can remove some redundancy by moving this into the while loop?

django_cloud_deploy/cli/prompt.py
```diff
@@ -32,260 +33,430 @@
...
-        """The default name to use."""
+    if 1 <= int(s) <= (len_options + 1):
+        return
+    else:
+        raise ValueError('Please enter a value between {} and {}'.format(
+            1, len_options + 1))
+
+
+def _binary_prompt(question: str, console: io.IO,
+                   default: Optional[str] = None) -> str:
```
brianquinlan: Would it make more sense to return a `bool`?

brianquinlan: -> Optional[str], no?

django_cloud_deploy/cli/prompt.py
```diff
@@ -32,260 +33,430 @@
...
+def _password_validate(s):
+    """Validates that a string is a valid password.
+
+    Args:
+        s: The string to validate.
+
+    Raises:
+        ValueError: if the input string is not valid.
+    """
+    if len(s) < 5:
```
```suggestion
    if len(s) < 6:
```

django_cloud_deploy/cli/prompt.py
```diff
@@ -32,260 +33,409 @@
...
+        pass
 
 
-class DjangoProjectNamePrompt(NamePrompt):
-    """Allow the user to enter a Django project name."""
+class TemplatePrompt(Prompt):
+    """Used as a base template for all Parameter Prompts interacting with user.
 
-    _PROMPT = 'Enter a Django project name, or leave blank to use'
+    They must have a prompt method that calls one of the _x_prompt functions.
```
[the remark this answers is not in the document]
brianquinlan: I don't think that you should resolve conversations until you've answered the questions.

django_cloud_deploy/cli/prompt.py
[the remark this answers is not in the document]
brianquinlan: quit is the usual way of handling this

django_cloud_deploy/cli/prompt.py
```diff
@@ -32,260 +33,430 @@
...
+    """Template for a simple string Prompt.
 
-    @classmethod
-    def _default_name(cls, arguments: Dict[str, Any]):
-        return 'home'
+    Any prompt that only needs to ask the user for a value without additional
+    branching or logic should derive from this class. Classes inheriting from
+    this should set the variables below.
+    """
+
```
brianquinlan: Could you explain what the variables mean, maybe with example values?

django_cloud_deploy/cli/prompt.py
```diff
@@ -32,260 +33,430 @@
...
+        """Handles the business logic and calls the prompts.
 
         Args:
-            s: The string to validate.
-            credentials: The OAuth2 Credentials object to use for api calls
-                during validation.
+            console: Object to use for user I/O.
+            step: Message to present to user regarding what step they are on.
+            args: Dictionary holding prompts answered by user and set up
+                arguments.
```
```suggestion
                command line arguments.
```

Please change everywhere.

django_cloud_deploy/cli/prompt.py
```diff
@@ -32,260 +33,430 @@
...
-        return 'mysite'
+    # Parameter must be set for dictionary key
+    PARAMETER = None
 
-    @staticmethod
-    def validate(s, credentials: Optional[credentials.Credentials] = None):
-        """Validates that a string is a valid Django project name.
+    def prompt(self, console: io.IO, step: str,
+               args: Dict[str, Any]) -> Dict[str, Any]:
+        """Handles the business logic and calls the prompts.
```
brianquinlan: You need fix fix 'Handles the business logic...'. That's basically the same as say 'Does what it needs to do...'

How about:

'Extracts user arguments through the command line or by displaying a prompt.'

enriquejosepadilla:
@brianquinlan If a conversation was resolved without the code being addressed then that was must have been a mistake. I apologize. Resolving it makes it easier for me to see what i have left to do.

django_cloud_deploy/cli/prompt.py
```diff
@@ -32,260 +33,441 @@
...
+    Typical usage:
+        # User can press enter if user doesn't want anything.
+        choice = _multiple_choice_prompt('Choose an option:\n{}\n',
+                                         ['Chicken', 'Salad', 'Burger'],
+                                         console,
+                                         default=None)
+
+    Returns:
+        The choice made by the user. Will be an index in the options list unless
+        default is None.
```
brianquinlan: Isn't that backwards? If the default is None then the returned value is guaranteed to be an index into the list. If it is not None then the value will either be an index into the list or it will be the provided default value.

enriquejosepadilla: you are correct.

django_cloud_deploy/cli/prompt.py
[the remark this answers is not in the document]
brianquinlan: OK

PR #179, opened 2019-01-15
Setup CircleCI

wangxf123456:


enriquejosepadilla:
whats credentials.json.enc?

wangxf123456:
> whats credentials.json.enc?

That is the service account key encrypted by a passphrase. We need to find some ways to let CircleCI access that file. Previously I decided to push this encrypted key and decrypt it at the beginning of integration test. Now I find another way to make CircleCI access this file, so credentials.json.enc is not used anymore. I will remove it from the pull request.

PR #183, opened 2019-01-17
Added user agent string to Project Client

enriquejosepadilla:
If we want to test if the user agent gets picked up this could be a good start.


enriquejosepadilla:
if http is not None and credentials is not None:
    raise ValueError('Arguments http and credentials are mutually exclusive.')

PR #184, opened 2019-01-17
User agent string to Project Client

enriquejosepadilla:


django_cloud_deploy/cloudlib/project.py
```diff
@@ -49,9 +51,11 @@ def __init__(self, cloudresourcemanager_service: discovery.Resource):
 
     @classmethod
     def from_credentials(cls, credentials: credentials.Credentials):
+        http_client = http.set_user_agent(http.build_http(), 'cloud-deploy')
```
brianquinlan: I think that 'django-cloud-deploy' might be better.

enriquejosepadilla:
> What is this about?

We want to collect metrics regarding the usage of our tool. This will allow us to be able to identify API calls made from our tool.

django_cloud_deploy/cloudlib/project.py
enriquejosepadilla: si

PR #196, opened 2019-01-28
Make crash handling work on windows

brianquinlan:


enriquejosepadilla:
this should throw errors on other OS right?

PR #215, opened 2019-02-06
Add test for crash_handling.

### LINKED ISSUE #195: Write tests for `crash_handling`

wangxf123456:
Fixes #195


django_cloud_deploy/tests/unit/crash_handling/crash_handling_test.py
```diff
@@ -0,0 +1,82 @@
...
+
+import unittest
+from unittest import mock
+from urllib import parse
+
+from django_cloud_deploy import crash_handling
+from django_cloud_deploy.cli import io
+
+
+@mock.patch('tempfile.mkstemp', return_value=(1, '/tmp/tmpfile'))
```
brianquinlan: This will will really get written to, right? Do you need to mock this?

django_cloud_deploy/tests/unit/crash_handling/crash_handling_test.py
```diff
@@ -0,0 +1,82 @@
...
+    def test_ignore_user_error(self, mock_open_browser, *unused_mocks):
+        test_io = io.TestIO()
+        test_io.answers.append('Y')  # Agree to open browser to create an issue
+        error = crash_handling.UserError('Test UserError')
+        with self.assertRaises(crash_handling.UserError):
+            try:
+                raise error
+            except KeyError as e:
+                crash_handling.handle_crash(e, 'command_fake', test_io)
+        mock_open_browser.assert_not_called()
```
brianquinlan: Can you assert that something was written to stderr?

django_cloud_deploy/tests/unit/crash_handling/crash_handling_test.py
wangxf123456: This will not. Because I also mocked os.fdopen, nothing will be written to this path. I was just worrying about creating a new file each time this test case is ran, so I mocked it.

django_cloud_deploy/tests/unit/crash_handling/crash_handling_test.py
wangxf123456: The logic of this test case was wrong. Now I assert KeyError is raised instead of asserting UserError is raised.

brianquinlan: But something should still be written to stderr, right?

enriquejosepadilla: Don't we assume if an error is thrown it has to show on the stderr?

brianquinlan: Yeah, I misunderstood how this worked.

PR #216, opened 2019-02-08
Reuse GCS bucket if it already exists.

### LINKED ISSUE #207: The tool should not crash if GCS bucket already exists

wangxf123456:
Fixes #207


enriquejosepadilla:
The tool attempts to create the bucket, and if it fails, it checks if the bucket already exists on the project. 

I'm curious about the possible solution of checking if bucket exists beforehand? The current solution will always make 2 API calls while checking will reduce it to one. (Not that we are low latency) 

I think this is fine but asking since this tends to be a common problem I encounter. 

@brianquinlan thoughts?

django_cloud_deploy/cloudlib/static_content_serve.py
```diff
@@ -44,6 +44,31 @@ def __init__(self, storage_service: discovery.Resource):
...
+            StaticContentServeError: When it fails to list buckets under the
+                given project.
+        """
+        request = self._storage_service.buckets().list(project=project_id)
+        response = request.execute()
+        if 'items' not in response:
+            raise StaticContentServeError(
+                'Unexpected response listing buckets in project "{}"'
+                ': {}'.format(project_id, response))
+        for item in response['items']:
```
brianquinlan: Just an FYI, you can write this as:

return any(item['name'] == bucketname for item in response['items'])

PR #217, opened 2019-02-08
Reuse service account

### LINKED ISSUE #208: The tool should not crash if service account already exists

wangxf123456:
Fixes #208


django_cloud_deploy/tests/integration/workflow_test.py
```diff
@@ -109,10 +109,13 @@ def test_create_service_account_key(self):
         member = 'serviceAccount:{}'.format(service_account_email)
         with self.delete_service_account(service_account_email):
             with self.reset_iam_policy(member, self.ROLES):
-                key_data = (self.service_account_workflow.
-                            create_service_account_and_key(
-                                self.project_id, service_account_id,
-                                'Test Service Account', self.ROLES))
+                # Assert no exceptions are raised when creating the same
+                # service account twice
+                for _ in range(2):
```
brianquinlan: This should be separate test.

wangxf123456: Done.

django_cloud_deploy/cloudlib/service_account.py
```diff
@@ -120,10 +120,12 @@ def create_service_account(self, project_id: str, service_account_id: str,
                     format(service_account_id, response))
         except errors.HttpError as e:
             if e.resp.status == 409:
-                raise ServiceAccountCreationError(
-                    'Service account {} already exists within project {}'.
-                    format(service_account_id, project_id))
-            elif e.resp.status == 400:
+                # A service account with the same name already exist. This is
```
enriquejosepadilla: I think its worth telling them that this is most likely because the tool was already run and has created these service accounts. Eg: 

Most likely our tool has created the service account already and is being re-run. This error code is fine as we just need the service accounts to exist. Later on we assign the appropriate roles for each one.

wangxf123456: Done. Thanks very much for rewording my comment!

PR #240, opened 2019-02-15
Add coverage report

enriquejosepadilla:


.circleci/config.yml
```diff
@@ -89,6 +89,18 @@ jobs:
...
+    docker:
+      - image: googleapis/nox:0.18.2
+
+    steps:
+      - checkout
+      - run:
+          name: Run Test
+          command: |
+            nox -f django_cloud_deploy/nox.py -s coverage
+          # Timeout of coverage report is set to 10 minutes
```
wangxf123456: By default CircleCI will wait 10 minutes for output. We can delete this limit here.

.circleci/config.yml
```diff
@@ -89,6 +89,18 @@ jobs:
             nox -f django_cloud_deploy/nox.py -s e2e_test
           # Timeout of e2e test is set to 30 minutes
           no_output_timeout: 30m
+  Coverage Report:
```
wangxf123456: You should add "Coverage Report" under "workflow". Otherwise you just defined this job and not running it.

enriquejosepadilla:
@wangxf123456 why is the unit test run on the 
    docker:
      - image: googleapis/nox:0.18.2

while the other ones are on the machine?

enriquejosepadilla:
@wangxf123456 Is there a way to make the coverage job be last? or add a hook to after integration and e2e finish ?

wangxf123456:
> @wangxf123456 why is the unit test run on the
> docker:
> - image: googleapis/nox:0.18.2
> 
> while the other ones are on the machine?

When you use an docker image in CircleCI, your tests will be run based on the docker image. Image "googleapis/nox:0.18.2 will install nox for you, while machine does not automatically provide nox. When using machines, you need to install it yourself. See https://github.com/GoogleCloudPlatform/django-cloud-deploy/blob/master/.circleci/config.yml#L62

CircleCI provides two kinds of environments to run your build. Docker image and machines. Integration tests and E2E tests will build and push docker images. But it is non-trivial about how to build a docker image inside a docker image. So I used machine for integration tests.

PR #252, opened 2019-02-21
Do not add duplicate bindings.

wangxf123456:
If the same role/member bindings exist multiple times, the policy update will fail.


django_cloud_deploy/cloudlib/static_content_serve.py
```diff
@@ -67,6 +67,26 @@ def _bucket_exist(self, project_id: str, bucket_name: str) -> bool:
                 ': {}'.format(project_id, response))
         return any(item['name'] == bucket_name for item in response['items'])
 
+    def _generate_updated_iam_policy(self, policy, member: str, role: str):
```
enriquejosepadilla: Docs please.

wangxf123456: Done.

PR #256, opened 2019-02-22
Backend var cleanup

### LINKED ISSUE #250: Save whether the project was deployed with GKE or GAE in the project's configuration file
If the project was deployed with GKE, and then we want to use "update" command to update it, because by default the "update" command uses GAE, it complains that the project cannot be deployed on GAE.

enriquejosepadilla:
fixes #250


django_cloud_deploy/workflow/__init__.py
```diff
@@ -279,8 +274,9 @@ def update_project(self,
         config_obj = config.Configuration(django_directory_path)
         project_id = config_obj.get('project_id')
         django_project_name = config_obj.get('django_project_name')
+        backend = config_obj.get('backend')
         cloud_sql_proxy_port = portpicker.pick_unused_port()
-        if not project_id or not django_project_name:
+        if not project_id or not backend or not cloud_sql_proxy_path:
```
wangxf123456: Why cloud_sql_proxy_path?

enriquejosepadilla: typo, good catch

PR #259, opened 2019-02-26
Refactor source generator.

wangxf123456:
This PR aims to simplify the logic of source generator while keep its existing logic.

Previously the logic is similar with the following:

```python
if generated():  # Whether our files are already generated
    return

if exist():   # Whether the directory already contains a requirements.txt if we want to generate requirements.txt
  generate_from_existing()   
else:
  generate_new()
```

The new logic is simple. "generate_new()" will assume the target directory is empty. And "generate_from_existing()" will assume the target directory contains the files we want to generate. So we can always call "generate_new" in "new" command and "generate_from_existing" in "cloudify" command.


django_cloud_deploy/skeleton/source_generator.py
```diff
@@ -30,41 +30,31 @@
 class _FileGenerator(object):  # pytype: disable=ignored-abstractmethod
```
enriquejosepadilla: should this look like:

class Prompt(abc.ABC):
    @abc.abstractmethod
    def foo():

wangxf123456: I removed @abstractmethd decorator for those methods. This is because those two methods can have variant parameters in subclasses.

PR #261, opened 2019-02-27
Enable deployment to different GAE services.

wangxf123456:
Different services on GAE will not interfere each other and can be deleted. This is helpful for GAE E2E test. #93


django_cloud_deploy/workflow/_deploygae.py
```diff
@@ -98,4 +99,12 @@ def deploy_gae_app(self,
             env=env_vars)
         if gcloud_result.returncode != 0:
             raise DeployNewAppError(gcloud_result.stderr)
-        return 'https://{}.appspot.com/'.format(project_id)
+
+        with open(app_yaml_path) as yaml_file:
+            attributes = yaml.load(yaml_file.read())
+        service_name = attributes.get('service')
+        if service_name:
```
enriquejosepadilla: Wont it always have a service name?

wangxf123456: Yes there will always be a service name. I modified the logic.

PR #264, opened 2019-02-27
Add prompts for cloudify command.

wangxf123456:


django_cloud_deploy/cli/prompt.py
```diff
@@ -924,6 +925,65 @@ def _validate(self, s: str):
                 ).format(s))
 
 
+class DjangoFilesystemPathCloudify(TemplatePrompt):
```
enriquejosepadilla: Did you check StringTemplatePrompt Class? This looks like it would be the ideal case for it.

django_cloud_deploy/cli/prompt.py
```diff
@@ -924,6 +925,65 @@ def _validate(self, s: str):
...
+
+        Returns:
+            A Copy of args + the new parameter collected.
+        """
+        new_args = copy.deepcopy(args)
+        if self._is_valid_passed_arg(console, step, args.get(self.PARAMETER),
+                                     self._validate):
+            return new_args
+
+        while True:
```
enriquejosepadilla: ask_prompt technically does this logic already.

django_cloud_deploy/cli/prompt.py
wangxf123456: I reimplemented this class with StringTemplatePrompt.

django_cloud_deploy/cli/prompt.py

PR #265, opened 2019-02-27
Modify settings file to get prepared for deploying existing apps.

wangxf123456:


django_cloud_deploy/skeleton/templates/settings_template/base_settings.py-tpl
```diff
@@ -42,7 +42,22 @@ MIDDLEWARE = [
 
 ROOT_URLCONF = '{{ project_name }}.urls'
 
-TEMPLATES = []
+TEMPLATES = [
+    {
+        # Add the templates required by CloudAdminSite.
+        'BACKEND': 'django.template.backends.django.DjangoTemplates',
+        'DIRS': [],
```
enriquejosepadilla: Why doesnt this DIRS have the 'DIRS': [os.path.join(BASE_DIR, 'templates')], ?

This might be easier to show me than have a conversation via comments.

enriquejosepadilla:
I sort of understand what change is happening, but I have no idea why this needed or how it affects deploying existing apps.

django_cloud_deploy/skeleton/templates/settings_template/base_settings.py-tpl
wangxf123456: "google_settings.py" contains setting changes needed for our app. For example, register cloud admin overwrite app and its dependencies.

We added "os.path.join(BASE_DIR, 'templates')" to 'DIRS in "google_settings.py", because this directory is needed by admin overwrite app. If we also put "os.path.join(BASE_DIR, 'templates')" here, then this directory is added twice.

When deploying existing apps, "base_settings.py" is not used. We will use their settings file as the base settings. Existing apps may or may not have "'DIRS': [os.path.join(BASE_DIR, 'templates')]". Their template directory could be somewhere else.  So we would like "google_settings.py" to handle this case.

brianquinlan: @enriquejosepadilla Please approve when you are satisfied with this change.

wangxf123456:
Actually this change comes from https://github.com/GoogleCloudPlatform/django-cloud-deploy/pull/137. I just made it a separate change.

PR #266, opened 2019-02-27
Modify new command to make it accept app engine service name parameter.

wangxf123456:
#93

### DIFF
--- a/django_cloud_deploy/cli/new.py
+++ b/django_cloud_deploy/cli/new.py
@@ -109,6 +109,12 @@ def add_arguments(parser):
         help=('Services necessary for the deployment. '
               'Test only, do not use.'))
 
+    parser.add_argument(
+        '--appengine-service-name',
+        dest='appengine_service_name',
+        nargs='+',
+        help=('App engine service name. Test only, do not use.'))
+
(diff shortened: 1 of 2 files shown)


django_cloud_deploy/cli/new.py
```diff
@@ -119,7 +125,8 @@ def main(args: argparse.Namespace, console: io.IO = io.ConsoleIO()):
         'project_creation_mode': workflow.ProjectCreationMode.CREATE,
         'bucket_name': getattr(args, 'bucket_name', None),
         'service_accounts': getattr(args, 'service_accounts', None),
-        'services': getattr(args, 'services', None)
+        'services': getattr(args, 'services', None),
```
enriquejosepadilla: where does services come from?

wangxf123456: I did not modify 'services'. I added 'appengine_service_name' and this can be provided by a SimpleNamespace object in e2e test. See https://github.com/GoogleCloudPlatform/django-cloud-deploy/blob/master/django_cloud_deploy/tests/e2e/gke_deploy_test.py#L107.

PR #274, opened 2019-03-07
Fix integration test about database reusing

wangxf123456:
The test did not fail before. This might be caused by some changes in GCP backend.

Now we will check whether the database already exist or not. The tool only try to create the database when that database does not exist.

### DIFF
--- a/django_cloud_deploy/cloudlib/database.py
+++ b/django_cloud_deploy/cloudlib/database.py
@@ -146,6 +146,11 @@ class DatabaseClient(object):
             project=project_id, instance=instance, database=database)
         try:
             response = request.execute()
+
+            # This means the database already exist. In this case we do not need
+            # to create the same database again.
+            if 'name' in response:
+                return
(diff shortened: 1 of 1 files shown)


django_cloud_deploy/cloudlib/database.py
```diff
@@ -142,6 +142,24 @@ def create_database_sync(self, project_id: str, instance: str,
         """
         # See:
         # https://cloud.google.com/sql/docs/mysql/admin-api/v1beta4/databases/insert
+        request = self._sqladmin_service.databases().get(
+            project=project_id, instance=instance, database=database)
+        try:
+            response = request.execute()
+
+            # This means the database already exist. In this case we do not need
```
brianquinlan: Move this outside of the `try` block.

wangxf123456: Done.

wangxf123456:
Will merge this PR without approval because it is blocking other PRs.

PR #275, opened 2019-03-08
GCP Resource Clean Up

### LINKED ISSUE #234: Automatic resource clean up of the GCP project we use for integration test
Some resources are not correctly deleted because of termination of tests when it is running (Ctrl+C), or the testing jobs are cancelled before it is finished. Resources are not deleted are left in the GCP project. We should find a way to automatically clean up those unused resources.

wangxf123456:
Sometimes GCP resources are not correctly cleaned up. There are several reasons causing this:

- The api call for resource deletion failed.
- The test halts before resource clean up finish. This can happen when something wrong with the CI environment or when we are running the test locally and press Ctrl+C before it finishes

Our billing account will be charged more for more resources, and the tests will be slower.

This PR added another job on presubmit, which automatically delete all resources more than 2 hours old. It should not affect the latest test runs because no tests will run for more than 2 hours.  

When making PRs, we do not care whether resource cleanup succeed or not. If any cleanups failed, then those resources could be cleaned up on other presubmits.

Fixes #234


django_cloud_deploy/tests/lib/test_base.py
```diff
@@ -100,16 +100,112 @@ def setUp(self):
...
+            project=self.project_id, instance=instance_name)
+        response = request.execute()
+        databases = [item['name'] for item in response['items']]
+        return databases
+
+
+class ResourceCleanUp(BaseTest):
     """Class for test cases which need resource cleaning up."""
 
+    def _delete_cluster(
```
wangxf123456: These changes are made because we want to reuse code for resource deletion in integration tests and the newly added resource cleanup.

django_cloud_deploy/tests/lib/test_base.py
```diff
@@ -100,16 +100,112 @@ def setUp(self):
...
-            database_name=self.database_name,
             image_tag=self.image_tag)
 
     def tearDown(self):
         shutil.rmtree(self.project_dir)
 
 
-class ResourceCleanUpTest(BaseTest):
+class ResourceList(BaseTest):
+    """Class for listing GCP resources.
```
wangxf123456: This class is used in integration test and resource cleanup.

django_cloud_deploy/tests/lib/test_base.py
```diff
@@ -87,8 +87,7 @@ def setUp(self):
         self.project_dir = tempfile.mkdtemp()
         image_name = utils.get_resource_name(resource_type='image')
         self.image_tag = '/'.join(['gcr.io', self.project_id, image_name])
-        self.instance_name = utils.get_resource_name(
-            resource_type='sql-instance')
+        self.instance_name = utils.get_resource_name(resource_type='instance')
```
wangxf123456: When parsing resource creation time, I used string.split('-'). The change here is to avoid unexpected results.

django_cloud_deploy/tests/e2e/gae_deploy_test.py
```diff
@@ -31,7 +31,7 @@
 from selenium.webdriver.chrome import options
 
 
-class GAEDeployAndUpdateE2ETest(test_base.ResourceCleanUpTest):
+class GAEDeployAndUpdateE2ETest(test_base.ResourceCleanUp):
```
wangxf123456: Renaming ResourceCleanUpTest to ResourceCleanUp, because this class is inherited by GCPResourceCleanUp, and it is not a test.

django_cloud_deploy/tests/cleanup/resource_cleanup.py
```diff
@@ -0,0 +1,113 @@
...
+
+class GCPResourceCleanUp(test_base.ResourceCleanUp, test_base.ResourceList):
+    """Clean up GCP resources more than 2 hours old."""
+
+    MAX_DIFF = datetime.timedelta(hours=2)
+
+    def _should_delete(self, resource_name: str) -> bool:
+        """Return whether the given resource should be deleted."""
+
+        # The resource created for testing should be in format similar with
```
brianquinlan: This comment should be part of the `args` section of the doc string.

django_cloud_deploy/tests/cleanup/resource_cleanup.py
```diff
@@ -0,0 +1,113 @@
...
+                if self._should_delete(service_account_name):
+                    binding['members'].remove(member)
+
+        policy['bindings'] = [b for b in policy['bindings'] if b['members']]
+        body = {'policy': policy}
+        request = cloudresourcemanager_service.projects().setIamPolicy(
+            resource=self.project_id, body=body)
+        request.execute()
+
+    def test_resource_cleanup(self):
```
brianquinlan:
```suggestion
    def cleanup_test_resources(self):
```

django_cloud_deploy/tests/lib/test_base.py
```diff
@@ -107,9 +107,111 @@ def tearDown(self):
...
+    This class is needed by resource cleanup and workflow test.
+    """
+
+    def list_service_accounts(self, service=None):
+        service = service or discovery.build(
+            'iam', 'v1', credentials=self.credentials, cache_discovery=False)
+        resource_name = '/'.join(['projects', self.project_id])
+        request = service.projects().serviceAccounts().list(name=resource_name)
+        accounts = []
+        while True:
```
brianquinlan: Maybe while `request`

django_cloud_deploy/tests/cleanup/resource_cleanup.py
wangxf123456: Done.

django_cloud_deploy/tests/cleanup/resource_cleanup.py
wangxf123456: Actually this whole class inherits unittest.TestCase indirectly. So to run this function, I can either make the function name starting with "test", or manually register this test case. Now I just manually register this test case.

django_cloud_deploy/tests/lib/test_base.py
wangxf123456: Done.

django_cloud_deploy/tests/cleanup/resource_cleanup.py
wangxf123456: I tried manually registering this test case and run it with pytest, looks like pytest cannot find it. I also tried run it with a script, for example, "python3 tests/cleanup/resource_cleanup.py", but nox will complain about not able to find module django-cloud-deploy. So what could be a good way to automatically run this script?

django_cloud_deploy/tests/cleanup/resource_cleanup.py
```diff
@@ -0,0 +1,126 @@
...
+        If a resource is not more than 2 hours old, we should not delete it
+        because these resources are still being used by other test runs.
+        Also, the GCP project we are using for test already contains some
+        resources. For example, service accounts for authentication. Those
+        resources should not be deleted.
+
+        Args:
+            resource_name: Name of the resource we want to check whether it
+                should be deleted. The resource created for testing should be
+                in format similar witt "db-20190307223549-b4wl".
```
enriquejosepadilla: typo witt?

django_cloud_deploy/tests/cleanup/resource_cleanup.py
```diff
@@ -0,0 +1,126 @@
...
+
+        policy['bindings'] = [b for b in policy['bindings'] if b['members']]
+        body = {'policy': policy}
+        request = cloudresourcemanager_service.projects().setIamPolicy(
+            resource=self.project_id, body=body)
+        request.execute()
+
+    def test_cleanup_resources(self):
+        """This is not a test, but cleans up test related resources."""
+        self.delete_expired_clusters()
```
enriquejosepadilla: If anyone of these throw an error wouldn all subsequent ones not happen? Even though this is not a test, is it either A) worth breaking up into individual "tests", or try catching each of these?

wangxf123456: Nice catch! Actually this is similar with what you just told me. The resource cleaning in e2e tests should not raise errors. I will implement this.

django_cloud_deploy/tests/cleanup/resource_cleanup.py
wangxf123456: Done

enriquejosepadilla:
Awesome job :+1:

PR #269, opened 2019-02-28
Retry when changing iam policies.

wangxf123456:
Fixes #268

PR #277, opened 2019-03-09
Fix retries for updating iam policy.

wangxf123456:
Previously I merged #269 for retrying iam policy change. But that change did not work. No matter how many times I retried, the api call always raise exception complaining about concurrent policy changes ([Link](https://circleci.com/gh/GoogleCloudPlatform/django-cloud-deploy/1315)). 

The reason is that when doing read-modify-write cycle to perform policy updates, the results returned by "get_iam_policy" contains an "[etag](https://cloud.google.com/iam/reference/rest/v1/Policy)" attribute. And when calling "set_iam_policy", I include the exactly the same etag in request. Whenever there are policy changes, a different etag will be generated. See [Link](http://googleapis.github.io/google-cloud-ruby/docs/google-cloud-storage/v1.8.0/Google/Cloud/Storage/Policy.html). This means there cannot be other iam policy changes between "get_iam_policy" and "set_iam_policy". Otherwise the server will refuse the updates. This is why it always fail even with retries.

The new retry will include "get_iam_policy". So even if there are new changes, we can still get the latest etag.


django_cloud_deploy/tests/unit/cloudlib/service_account_test.py
```diff
@@ -295,10 +295,13 @@ def test_create_service_account_key_failure(self):
     def test_request_with_retry_simple_success(self):
         responses = [1]
```
enriquejosepadilla: Dont use magic number.

django_cloud_deploy/tests/unit/cloudlib/service_account_test.py
```diff
@@ -295,10 +295,13 @@ def test_create_service_account_key_failure(self):
...
-        response = service_account.ServiceAccountClient._request_with_retry(
-            request)
-        self.assertEqual(request.call_count, 1)
-        self.assertEqual(response, 1)
+        with mock.patch(
+                __name__ + '.ProjectsFake.setIamPolicy', return_value=request):
+            response = (
+                self._service_account_client._update_iam_policy_with_retry(
+                    PROJECT_ID, FAKE_SERVICE_ACCOUNT, [FAKE_ROLE]))
+            self.assertEqual(request.call_count, 1)
```
enriquejosepadilla: Dont use magic number.

django_cloud_deploy/tests/unit/cloudlib/service_account_test.py
```diff
@@ -307,18 +310,24 @@ def test_request_with_retry_success_at_second_time(self):
...
-        response = service_account.ServiceAccountClient._request_with_retry(
-            request)
-        self.assertEqual(request.call_count, 2)
-        self.assertEqual(response, 1)
+        with mock.patch(
+                __name__ + '.ProjectsFake.setIamPolicy', return_value=request):
+            response = (
+                self._service_account_client._update_iam_policy_with_retry(
+                    PROJECT_ID, FAKE_SERVICE_ACCOUNT, [FAKE_ROLE]))
+            self.assertEqual(request.call_count, 2)
```
enriquejosepadilla: Dont use magic number.

Issue #205, opened 2019-02-05
django-cloud-deploy new --use-existing-project crashes when the user doesn't have permission to deploy


josiahdahl:
## Issue running command
django-cloud-deploy new

## Please describe what is wrong
`django-cloud-deploy new --use-existing-project` crashes when the user doesn't have permission to deploy

## Environment information

Django Cloud Deploy: [0.1.5]

Platform: [Linux-4.15.0-43-generic-x86_64-with-Ubuntu-18.04-bionic]

Gcloud Version: [233.0.0]

Cloud SQL Proxy Version: [Cloud SQL Proxy: version 1.13; sha ac834ce67862b7c4ed8bdbe86e664572f820c245 built Tue Oct  9 23:39:16 UTC 2018]

Python Version: [3.6.7 (default, Oct 22 2018, 11:32:17)  [GCC 8.2.0]]

Docker Version: [Docker version 18.09.0, build 4d60db4]

## Traceback
```.pytb
Traceback (most recent call last):
  File "/home/jdahl/Code/scratch/django-deploy/lib/python3.6/site-packages/pexpect/spawnbase.py", line 166, in read_nonblocking
    s = os.read(self.child_fd, size)
OSError: [Errno 5] Input/output error

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/jdahl/Code/scratch/django-deploy/lib/python3.6/site-packages/pexpect/expect.py", line 111, in expect_loop
    incoming = spawn.read_nonblocking(spawn.maxread, timeout)
  File "/home/jdahl/Code/scratch/django-deploy/lib/python3.6/site-packages/pexpect/pty_spawn.py", line 485, in read_nonblocking
    return super(spawn, self).read_nonblocking(size)
  File "/home/jdahl/Code/scratch/django-deploy/lib/python3.6/site-packages/pexpect/spawnbase.py", line 171, in read_nonblocking
    raise EOF('End Of File (EOF). Exception style platform.')
pexpect.exceptions.EOF: End Of File (EOF). Exception style platform.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/jdahl/Code/scratch/django-deploy/lib/python3.6/site-packages/django_cloud_deploy/workflow/_deploygae.py", line 63, in deploy_gae_app
    process.expect('Do you want to continue (Y/n)?')
[...]

PR #286, opened 2019-03-11
Checks permission for existing projects.

enriquejosepadilla:
Fixes: https://github.com/GoogleCloudPlatform/django-cloud-deploy/issues/205

User study would provide users with projects they are not owner of GAE project, or they were not editors for GKE. This checks to make sure and crashes early.

Only applies to existing-projects.

### DIFF
--- a/django_cloud_deploy/cli/prompt.py
+++ b/django_cloud_deploy/cli/prompt.py
@@ -631,7 +631,8 @@ class GoogleExistingProjectId(TemplatePrompt):
         # Regex will catch user:email@email.com
         # and serviceAccount:email@email.com
         # which are currently the only cases
-        active_account = re.escape(r'\w+:{}'.format(active_account))
+        active_account = re.escape(active_account)
+        active_account = r'\w+:{}'.format(active_account)
         if re.search(active_account, ' '.join(owners)):
             return True
 
--- a/django_cloud_deploy/cli/prompt.py
+++ b/django_cloud_deploy/cli/prompt.py
@@ -613,7 +613,7 @@ class GoogleExistingProjectId(TemplatePrompt):
             active_account: Account that is logged in on gcloud cli.
         """
         # The user must have logged in
-        assert active_account != ''
+        assert active_account not in ['', None], "User must log in via gcloud"
 
         permissions = self.project_client.get_project_permissions(project_id)
         owner_permission = list(
@@ -628,7 +628,10 @@ class GoogleExistingProjectId(TemplatePrompt):
         if editor_permission:
             editors = editor_permission[0].get('members', [])
 
-        active_account = re.escape(active_account)
+        # Regex will catch user:email@email.com
+        # and serviceAccount:email@email.com
+        # which are currently the only cases
+        active_account = re.escape(r'\w+:{}'.format(active_account))
         if re.search(active_account, ' '.join(owners)):
             return True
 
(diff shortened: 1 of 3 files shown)


wangxf123456:
If the integration tests fail, you can merge with the master branch and try again.

enriquejosepadilla:
the change from copy.deepcopy to dict(args) was due to a very wierd error arising when trying to run locally. As @brianquinlan had suggested using this approach (downside is that it is shallow), i think for future debugging is better than future possibility of nested args.

django_cloud_deploy/cli/prompt.py
```diff
@@ -541,50 +543,102 @@ class GoogleExistingProjectId(TemplatePrompt):
...
+
+        owners = []
+        editors = []
+        if owner_permission:
+            owners = owner_permission[0].get('members', [])
+        if editor_permission:
+            editors = editor_permission[0].get('members', [])
+
+        active_account = re.escape(active_account)
+        if re.search(active_account, ' '.join(owners)):
```
wangxf123456: For example, my account is "tim@gmail.com", and this account does not exist in either owners or editors. And there is another account called "blablatim@gmail.com", who is the owner. Then re.search will return true when active_account is "tim@gmail.com", but it is not necessarily an owner.

django_cloud_deploy/cli/prompt.py
```diff
@@ -1160,9 +1215,12 @@ def prompt(self, command: Command, console: io.IO,
         step_template = '<b>[{}/{}]</b>'
         first_step = step_template.format(1, total_steps)
 
-        creds = self._get_creds(console, first_step, args)
+        auth_client = auth.AuthClient()
+        creds = self._get_creds(console, first_step, args, auth_client)
+        active_account = auth_client.get_active_account()
```
wangxf123456: auth_client.get_active_account() will return an empty string or None (I am not sure whether it is None or empty string in python) if the user has not logged in with gcloud. Will this be a problem?

django_cloud_deploy/tests/unit/cli/prompt_test.py
```diff
@@ -286,6 +303,25 @@ def test_existing_prompt(self, unused_mock):
...
+        test_io = io.TestIO()
+        args = {
+            'backend': 'gae',
+            'project_creation_mode': workflow.ProjectCreationMode.MUST_EXIST,
+            'project_id': 'projectid-123',
+            'use_existing_project': True
+        }
+        test_io.answers.append('projectid-123')
+
+        with self.assertRaises(SystemExit):
```
wangxf123456: Why SystemExit?

django_cloud_deploy/cli/prompt.py
enriquejosepadilla: it returns an empty string I confirmed this manually. but I can add a check to make sure it doesnt crash.

django_cloud_deploy/cli/prompt.py
enriquejosepadilla: that is a good catch.

django_cloud_deploy/tests/unit/cli/prompt_test.py
enriquejosepadilla: if the user passes it as a flag it does a system exit. I forgot to add a test so it just pops out a new answer in the append.

django_cloud_deploy/cli/prompt.py
enriquejosepadilla: There was a check already asserting it is not an empty string, as the user must have been logged in by that point.

PR #297, opened 2019-03-13
Ask user whether they would like to take a survey at the end of deployment process.

wangxf123456:
I did not add this at the end of update command. Update command will be used several times. It would be annoying to ask this question every time the user uses update command.


enriquejosepadilla:
I agree that asking after every update would be annoying, but we lose some metrics in regard of continual use of our tool. Maybe ask after the 5th / 10th time? @brianquinlan for thoughts.

django_cloud_deploy/cli/prompt.py
```diff
@@ -1075,6 +1075,31 @@ def _validate(self, s: str):
...
+            console: Object to use for user I/O.
+            step: Message to present to user regarding what step they are on.
+            args: Dictionary holding prompts answered by user and set up
+                command-line arguments.
+        """
+        msg = ('Would you like to take a survey to provide your feedback for '
+               'the deployment process? [Y/n]')
+
+        do_survey = _binary_prompt(msg, console, default='Y')
+        if do_survey:
```
brianquinlan: Could you move this to new.py

django_cloud_deploy/cli/prompt.py
```diff
@@ -1075,6 +1075,31 @@ def _validate(self, s: str):
...
+        Args:
+            console: Object to use for user I/O.
+            step: Message to present to user regarding what step they are on.
+            args: Dictionary holding prompts answered by user and set up
+                command-line arguments.
+        """
+        msg = ('Would you like to take a survey to provide your feedback for '
+               'the deployment process? [Y/n]')
+
+        do_survey = _binary_prompt(msg, console, default='Y')
```
```suggestion
        do_survey = _binary_prompt(msg, console, default='N')
```

django_cloud_deploy/cli/prompt.py
```diff
@@ -1075,6 +1075,31 @@ def _validate(self, s: str):
...
+        their web browser and direct them to an existing Qualtics survery.
+
+        Args:
+            console: Object to use for user I/O.
+            step: Message to present to user regarding what step they are on.
+            args: Dictionary holding prompts answered by user and set up
+                command-line arguments.
+        """
+        msg = ('Would you like to take a survey to provide your feedback for '
+               'the deployment process? [Y/n]')
```
```suggestion
               'the deployment process? [y/N]')
```

django_cloud_deploy/cli/prompt.py
```diff
@@ -1075,6 +1075,31 @@ def _validate(self, s: str):
...
+
+    def prompt(self, console: io.IO):
+        """Ask users to take a survey.
+
+        If the user would like to take the survey, then the method will open
+        their web browser and direct them to an existing Qualtics survery.
+
+        Args:
+            console: Object to use for user I/O.
+            step: Message to present to user regarding what step they are on.
```
brianquinlan: You don't have any of these...does this class actually follow the contract for being a prompt? It might make sense for this to not be a prompt?

django_cloud_deploy/cli/prompt.py

django_cloud_deploy/cli/prompt.py

django_cloud_deploy/cli/prompt.py
wangxf123456: Done

django_cloud_deploy/cli/prompt.py
wangxf123456: I made this class not inheriting TemplatePrompt

django_cloud_deploy/cli/prompt.py
```diff
@@ -1136,6 +1136,33 @@ def _validate(self, s: str):
                               '"test@example.com"').format(s))
 
 
+class TakeSurveyPrompt(object):
```
brianquinlan: OK, maybe it would make sense at this point to put this in it's own file, maybe with the logic to open the web browser?

wangxf123456: It is using the private method "_binary_prompt" in prompt.py. So maybe also make that method public?

brianquinlan: Ahhh....yes, you could make that public. I think that it make sense to put this in a new file because I can imagine wanting to do the survey for `update` to but maybe only the first time or something.

django_cloud_deploy/utils/survey.py
```diff
@@ -0,0 +1,39 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Functions for asking users to take a survey at the end of deployment."""
+
+from django_cloud_deploy.cli import io
```
brianquinlan: Well, now you may as well open the link at the end (since this isn't in the prompts module anymore). Sorry about that.

wangxf123456: Done.

django_cloud_deploy/cli/prompt.py
wangxf123456: I think I will add survey prompt for update on a follow up PR. Right now I have not figured out how to prompt only once.

django_cloud_deploy/utils/survey.py
```diff
@@ -0,0 +1,42 @@
...
+def prompt_for_survey(console: io.IO) -> bool:
+    """Ask users to take a survey.
+
+    If the user would like to take the survey, then the method will open
+    their web browser and direct them to an existing Qualtrics survey.
+
+    Args:
+        console: Object to use for user I/O.
+
+    Returns:
```
brianquinlan: It actually doesn't return anything

django_cloud_deploy/utils/survey.py
```diff
@@ -0,0 +1,42 @@
...
+
+from django_cloud_deploy.cli import io
+from django_cloud_deploy.cli import prompt
+from django_cloud_deploy.utils import webbrowser
+
+
+_SURVEY_LINK = 'https://google.qualtrics.com/jfe/form/SV_3wwUubKBJnC7Fxr'
+
+
+def prompt_for_survey(console: io.IO) -> bool:
```
brianquinlan:
```suggestion
def prompt_for_survey(console: io.IO):
```

django_cloud_deploy/utils/survey.py
wangxf123456: My bad.

django_cloud_deploy/utils/survey.py
wangxf123456: Done

PR #298, opened 2019-03-13
Do not do components update after installing gcloud.

wangxf123456:
Sometimes this causes installation to timeout on CircleCI. And the version of gcloud cli we are installing on CircleCI is 236.0.0. This one is new enough to run e2e test.


enriquejosepadilla:
Since we use gcloud for deploying GAE we have to be careful with this in the future.

PR #306, opened 2019-03-14
Wait for app engine update ready.

wangxf123456:
Sometimes the test make request to the service before app engine put traffic to the most recent service. This causes the test to fail.


django_cloud_deploy/tests/e2e/gae_deploy_test.py
```diff
@@ -57,6 +58,28 @@ def tearDown(self):
     def _get_with_retry(url: str) -> requests.models.Response:
         return requests.get(url)
 
+    @backoff.on_predicate(backoff.expo, interval=2.0, logger=None, max_tries=5)
+    def _is_latest_app_ready(self) -> float:
```
enriquejosepadilla: return type is now bool

PR #311, opened 2019-03-15
Remove upper bound constraint on urllib3

### LINKED ISSUE #310: Support newer versionn of urllib3
django-cloud-deploy 0.1.7 has requirement urllib3==1.23, but you'll have urllib3 1.24.1 which is incompatible.

Maybe `urllib3>=1.23` instead?

wangxf123456:
The incompatibility between urllib3 1.24 and requests is resolved. See https://github.com/requests/requests/issues/4830

Fixes #310


setup.py
```diff
@@ -28,7 +28,7 @@
 
 
 install_requires = [
-    'urllib3==1.23',  # https://github.com/requests/requests/issues/4830
+    'urllib3>=1.23',
```
brianquinlan: You could add <2 if you want to assume that they are fallowing semver.

wangxf123456: Done.

PR #314, opened 2019-03-18
Add a function to guess settings file path of an existing Django project

wangxf123456:
We will prompt user to provide their settings file path when deploying existing django project. This function will be used to provide a default value for that prompt.


django_cloud_deploy/skeleton/utils.py
```diff
@@ -72,3 +74,58 @@ def is_valid_django_project(django_directory_path: str) -> bool:
...
+        django_directory_path: Absolute path of a Django project.
+
+    Returns:
+        Absolute path of settings.py of the given Django project. If cannot find
+        it, return an empty string.
+    """
+
+    manage_py_path = os.path.join(django_directory_path, 'manage.py')
+    if not os.path.exists(manage_py_path):
+        return ''
```
brianquinlan: Maybe return None in these cases?

wangxf123456: Done.

django_cloud_deploy/skeleton/utils.py
```diff
@@ -72,3 +76,58 @@ def is_valid_django_project(django_directory_path: str) -> bool:
...
+        2. Find settings files containing "prod" in the same directory with
+           local settings file. If this file is found, return it. If not, return
+           local settings file.
+        3. If manage.py does not exist or cannot find local settings file path
+           from manage.py, return an empty string.
+    Args:
+        django_directory_path: Absolute path of a Django project.
+
+    Returns:
+        Absolute path of settings.py of the given Django project. If cannot find
```
brianquinlan:
```suggestion
        Absolute path of settings.py of the given Django project. If it cannot be found, return None
```

PR #316, opened 2019-03-19
Add prompt for settings file of existing Django project

wangxf123456:


django_cloud_deploy/cli/prompt.py
```diff
@@ -1136,6 +1138,81 @@ def _validate(self, s: str):
                               '"test@example.com"').format(s))
 
 
+class DjangoSettingsPathPrompt(StringTemplatePrompt):
+    """Allow the user to enter the settings file path of a Django project."""
+
+    PARAMETER = 'django_settings_path'
+    MESSAGE = ('{} Enter the settings file path you want to use for deployment')
```
brianquinlan: This isn't completely clear to me. Maybe explain it and we can work on the wording a bi t?

django_cloud_deploy/cli/prompt.py
```diff
@@ -1136,6 +1138,81 @@ def _validate(self, s: str):
...
+
+        if not django_settings_path.endswith('.py'):
+            raise ValueError(
+                '["{}"] is not a .py file.'.format(django_settings_path))
+
+        module_relative_path = os.path.relpath(django_settings_path,
+                                               project_dir)
+
+        # Remove ".py" at the end.
+        module_name = module_relative_path[:-3].replace('/', '.')
```
brianquinlan: os.path.splitext will remove the .py for you - and you can use it for the extension checking above too

django_cloud_deploy/cli/prompt.py
```diff
@@ -1136,6 +1138,81 @@ def _validate(self, s: str):
...
+            args: Dictionary holding prompts answered by user and set up
+                command-line arguments.
+
+        Returns:
+            A Copy of args + the new parameter collected.
+        """
+        new_args = dict(args)
+
+        django_directory_path = args.get('django_directory_path', None)
+        validate = functools.partial(self._validate, django_directory_path)
```
brianquinlan: Move this right before _ask_prompt

wangxf123456: Actually it is used on the next line
```python
if self._is_valid_passed_arg(console, step, args.get(self.PARAMETER, None), validate):
```

django_cloud_deploy/cli/prompt.py

django_cloud_deploy/cli/prompt.py
brianquinlan: Ahhh, thanks.

django_cloud_deploy/cli/prompt.py
wangxf123456: I added more comments for this class.

brianquinlan: I mean that the message to the user isn't completely clear. I can't think of anything much better:

```suggestion
    MESSAGE = ('{} Enter the path of the Django settings file that should be used for deployment')
```

django_cloud_deploy/cli/prompt.py
```diff
@@ -1136,6 +1138,89 @@ def _validate(self, s: str):
...
+
+        Returns:
+            A Copy of args + the new parameter collected.
+        """
+        new_args = dict(args)
+
+        django_directory_path = args.get('django_directory_path_cloudify', None)
+        validate = functools.partial(self._validate, django_directory_path)
+        if self._is_valid_passed_arg(console, step,
+                                     args.get(self.PARAMETER, None), validate):
```
```suggestion
                                     args.get(self.PARAMETER), validate):
```

PR #319, opened 2019-03-20
Fix webbrowser.open_url()

wangxf123456:
Add restoration of fds after webbrowser.open().

I found this problem because I added survey prompt after webbrowser.open_url(), but I cannot see the output of the prompt. But if I press enter, my web browser will take me to the survey page.


django_cloud_deploy/utils/webbrowser.py
```diff
@@ -27,6 +27,16 @@ def open_url(url: str):
...
     """
+
+    # Save previous standard file descriptors
+    prev_stderr_fd = os.dup(2)
+    prev_stdout_fd = os.dup(1)
     with open(os.devnull, 'wb') as f:
+        # redirect stderr and stdout to os.devnull
         os.dup2(f.fileno(), 2)
+        os.dup2(f.fileno(), 1)
         webbrowser.open(url)
```
brianquinlan: Should you use try/finally in case webbrowser.open fails?

PR #321, opened 2019-03-22
Add prompt for path of requirements.txt

wangxf123456:
When deploying an existing Django project, sometimes it is hard to guess the path of their requirements.txt. We need to prompt user for this information.


enriquejosepadilla:
Isn't the requirements.txt file in the base of the project path?

wangxf123456:
Maybe not. For example, https://github.com/django/djangoproject.com.

enriquejosepadilla:
1. I believe it needs to be added to cloudfy_prompt_order ?
2. It uses utils.guess_requirements_path , Would it make sense to use that value for _is_valid_passed_arg to streamline the process?

wangxf123456:
1. Yeah. There is still some changes to be made to source_generator.py to make everything work. Maybe I can add this on the next PR.
2. Done

PR #327, opened 2019-03-26
Remove packages not required by our tool.

wangxf123456:
'google-cloud-storage', 'psycopg2-binary' and 'google-cloud-logging' are not required by our tool. They are required by the Django project generated by our tool. We should not mark them as the dependency of our tool.

Our tool needs 'absl-py' for testing. But it is not listed in setup.py. The testing dependencies can be installed by 
```
pip install .[test]
```


django_cloud_deploy/skeleton/source_generator.py
```diff
@@ -694,13 +694,15 @@ def install_requirements(self, project_dir: str):
...
         try:
-            subprocess.check_call(
+            subprocess.call(
                 ['python3', '-m', 'pip', 'install', '-r', requirements_path],
                 stdout=subprocess.DEVNULL,
                 stderr=subprocess.DEVNULL)
         except (SystemExit, subprocess.CalledProcessError) as e:
-            raise crash_handling.UserError(
-                'Not able to install Django project dependencies.') from e
+            print(('Fail to install some packages listed in {}. This may or '
```
wangxf123456: This change is made because our tool will install all the packages, but some of them may not be required for deployment. For example, "mysqlclient". When installing "mysqlclient" on CircleCI, we will see ```EnvironmentError: mysql_config not found```. But actually "mysqlclient" is not required for deployment.

enriquejosepadilla: :s/fail/failed

enriquejosepadilla: the errors. -> any errors

wangxf123456: Done

PR #348, opened 2019-04-03
Use instace class F2 for projects deployed on GAE.

wangxf123456:
By default GAE uses instance class F1. See [link](https://cloud.google.com/appengine/docs/standard/#instance_classes). This is not enough for most of the existing Django projects.

Also deleted configurations about static files in app.yaml. We are not using it for static content serving.


enriquejosepadilla:
how do you know that it is not enough?

wangxf123456:
After I go to the website, I see errors like
```
Receives error “Exceeded soft memory limit of 128 MB with 157 MB after servicing 0 requests total. Consider setting a larger instance class in app.yaml.”, 
```

enriquejosepadilla:
worth putting that in the description jajaja :) Approved.

PR #359, opened 2019-04-08
Only allow open billing accounts

### LINKED ISSUE #341: Fix billing open account
Currently you are able to avoid passing in a correct billing account name if you look at the names provided and choose any of them even if they are not usable. (Easy to do when debugging the billing_accounts found in the validate function)

enriquejosepadilla:
fix #341

### DIFF
--- a/django_cloud_deploy/cli/prompt.py
+++ b/django_cloud_deploy/cli/prompt.py
@@ -831,10 +831,7 @@ class BillingPrompt(TemplatePrompt):
         """
 
         billing_accounts = self.billing_client.list_billing_accounts(only_open_accounts=True)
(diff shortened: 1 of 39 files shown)


enriquejosepadilla:
waiting for lint PR

django_cloud_deploy/cli/prompt.py
```diff
@@ -828,9 +828,10 @@ def _validate(self, s):
         billing_accounts = self.billing_client.list_billing_accounts()
```
wangxf123456:
```suggestion
        billing_accounts = self.billing_client.list_billing_accounts(only_open_accounts=True)
```

PR #364, opened 2019-04-10
Do not use unnecessary constraints for Django version.

wangxf123456:
1. Do not use unnecessary constraints for Django version. This will make the tool support less existing projects.

2. When generating "requirements-google.txt", if it is an existing project, we do not list "Django" in it. This is because we expect the existing project is valid, it must explicitly or implicitly have "Django" as a dependency. We do not need to list "Django" as dependency again.


wangxf123456:
@enriquejosepadilla PTAL. Because #377 is merged, I do not think we are using any Django 2.0 only features.

PR #380, opened 2019-04-19
Allow customized location for static files.

wangxf123456:
settings.STATIC_ROOT is where "python manage.py collectstatic" put static files.


django_cloud_deploy/workflow/__init__.py
```diff
@@ -266,6 +266,7 @@ def create_and_deploy_new_project(
             self._enable_service_workflow.enable_required_services(
                 project_id, required_services)
 
+        static_content_dir = settings.STATIC_ROOT
```
enriquejosepadilla: I don't understand how this is equivalent to static_content_dir = os.path.join(django_directory_path, 'static') 

Couldn't django directory path be in another location than the current directory? I thought python manage.py collectstatic is for the current directory.

PR #382, opened 2019-04-19
Fix test for cloudify using GKE

wangxf123456:


enriquejosepadilla:
This the one that has been making tests fail correct?

wangxf123456:
> This the one that has been making tests fail correct?

Yes

PR #385, opened 2019-04-22
Do not use latest version of urllib3

wangxf123456:
Urllib3 did a release this morning for version 1.25.

"requests" is one of our dependency, which has dependency "[urllib3<1.25](https://github.com/kennethreitz/requests/blob/master/setup.py#L47)". This will cause our tool fail to run.


setup.py
```diff
@@ -28,7 +28,7 @@
 
 
 install_requires = [
-    'urllib3>=1.23,<2',
+    'urllib3>=1.23,<1.25',
```
brianquinlan: Maybe add a comment explaining why?

wangxf123456: Sure.

PR #392, opened 2019-05-14
Make database client able to create a new database user.

wangxf123456:
We will not use the default database user "postgres" and ask user for a username.

I will add a test case after we change the whole workflow.


django_cloud_deploy/tests/integration/cloudlib_test.py
```diff
@@ -83,26 +83,6 @@ def test_create_duplicate_service_account(self):
                 self._ROLES)
 
 
-class DatabaseClientIntegrationTest(test_base.DjangoFileGeneratorTest,
```
enriquejosepadilla: Why was this removed?

PR #394, opened 2019-05-15
Make DCD support Django mangement commands related to database

wangxf123456:
Now users can do the following:
```
(django_project)$ django-cloud-deploy cloudify
...
Your app is running at https://django-123456.appspot.com/


(django_project)$ django-cloud-deploy manage migrate
Do migration on cloud sql database

(django_project)$ django-cloud-deploy manage createsuperuser
Create superuser on cloud sql database

(django_project)$ django-cloud-deploy manage migrate my_app zero
Reverse all migrations for an app
```

I am not sure whether this is valuable or not. But since it is not very hard to add this functionality, we can add it. It is fine that nobody uses this functionality.


django_cloud_deploy/cloudlib/database.py
```diff
@@ -241,7 +241,6 @@ def with_cloud_sql_proxy(self,
         Raises:
             DatabaseError: If cloud sql proxy failed to start after 5 seconds.
         """
-        db.close_old_connections()
```
wangxf123456: This line is deleted, because this line requires the Django environment to be set up. But when running management commands, we do not setup Django environment.

PR #398, opened 2019-05-17
Fix user agent string

enriquejosepadilla:
We previosuly thought we were sending the user agent string, but https://github.com/googleapis/google-api-python-client/issues/656 showed us that it was implemented incorrectly. The additional code from utils is required until this 
PR: https://github.com/googleapis/google-api-python-client/pull/658 is merged.


django_cloud_deploy/cloudlib/_utils.py
```diff
@@ -0,0 +1,62 @@
+# Copyright 2018 Google LLC
```
wangxf123456:
```suggestion
# Copyright 2019 Google LLC
```

django_cloud_deploy/cloudlib/_utils.py
```diff
@@ -0,0 +1,62 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Helper file for common functionality within cloublib."""
+
+
+def set_user_agent(http, user_agent):
+    """Set the user-agent on every request.
+
+    Args:
+       http - An instance of google_auth_httplib2.AuthorizedHttp.
```
brianquinlan: You can just use type annotations to indicate this.

django_cloud_deploy/cloudlib/_utils.py
```diff
@@ -0,0 +1,62 @@
...
+       user_agent: string, the value for the user-agent header.
+
+    Returns:
+       A modified instance of http that was passed in.
```
brianquinlan: This is a convention in Python that mutation methods do *not* return what the mutate.

For example:
`sort()` does a mutation and returns ` None` but `sorted()` returns a copy and returns that copy.

django_cloud_deploy/cloudlib/_utils.py
```diff
@@ -0,0 +1,62 @@
...
+            headers['user-agent'] = user_agent
+        resp, content = request_orig(uri,
+                                     method,
+                                     body,
+                                     headers,
+                                     redirections=redirections,
+                                     connection_type=connection_type)
+        return resp, content
+
+    http.request = new_request
```
brianquinlan: Wait, why can't subclass here?

enriquejosepadilla: subclass google_auth_httplib2.AuthorizedHttp?

django_cloud_deploy/cloudlib/_utils.py
enriquejosepadilla: ack

django_cloud_deploy/cloudlib/_utils.py
enriquejosepadilla: ack

django_cloud_deploy/cloudlib/_utils.py
enriquejosepadilla: ack

django_cloud_deploy/cloudlib/_utils.py
brianquinlan: Yeah. That seems like the obvious way to customize a classes behavior and I was wondering if that would be more appropriate here.

enriquejosepadilla:
The [newest version](https://github.com/googleapis/google-api-python-client/pull/663) of google-api-python client fixes the bug that we encountered when trying to use this. Updated the package versioning as well as removed duplicate code. 

Latest output of print statements:

------------Create Project-------
{'content-type': 'application/json', 'content-length': '108', 'accept': 'application/json', 'accept-encoding': 'gzip, deflate', 'user-agent': 'django-cloud-deploy/0.3.4 google-api-python-client/1.7.9 (gzip)'}

Which shows that the user agent string has been appeneded.

django_cloud_deploy/cloudlib/project.py
```diff
@@ -50,10 +51,9 @@ def __init__(self, cloudresourcemanager_service: discovery.Resource):
 
     @classmethod
     def from_credentials(cls, credentials: credentials.Credentials):
-        http_client = http.set_user_agent(http.build_http(),
-                                          'django-cloud-deploy')
-        auth_http = google_auth_httplib2.AuthorizedHttp(credentials,
-                                                        http=http_client)
+        auth_http = google_auth_httplib2.AuthorizedHttp(credentials)
+        user_agent = '/'.join(['django-cloud-deploy', __version__.__version__])
+        auth_http = http.set_user_agent(auth_http, user_agent)
```
brianquinlan:
```suggestion
        http.set_user_agent(auth_http, user_agent)
```

